PR #8, opened 2023-05-25
Add ModelMesh custom status

ruivieira:


tteofili (approved):

api/v1alpha1/trustyaiservice_types.go
```diff
@@ -39,9 +47,6 @@ type MetricsSpec struct {
 
 // TrustyAIServiceSpec defines the desired state of TrustyAIService
 type TrustyAIServiceSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
 	// The namespace in which to deploy the image
 	Namespace string `json:"namespace"`
```
danielezonca, line 51: Sorry, it is not related with the PR but I just noticed it. I think this option is not necessary and actually it is error prone: I expect user to create this resource _in_ the namespace where they want to deploy TrustyAI Service.
This simplifies the model and also the permission part (the operator just need to have access to the namespace where the resource is).
Moreover the namespace where the operator is going to run, it should be "restricted" to the user so we cannot assume the CR is created in that namespace by the user

ruivieira: I'll do a separate PR for it. Tracked with https://github.com/trustyai-explainability/trustyai-service-operator/issues/11

controllers/trustyaiservice_controller.go
```diff
@@ -419,3 +427,64 @@ func (r *TrustyAIServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&trustyaiopendatahubiov1alpha1.TrustyAIService{}).
 		Complete(r)
 }
+
+func (r *TrustyAIServiceReconciler) reconcileStatuses(instance *trustyaiopendatahubiov1alpha1.TrustyAIService, ctx context.Context) error {
+	deploymentList := &appsv1.DeploymentList{}
+	labelSelector := client.MatchingLabels{"app.kubernetes.io/name": "modelmesh-controller"}
+	err := r.Client.List(ctx, deploymentList, client.InNamespace("modelmesh-serving"), labelSelector)
```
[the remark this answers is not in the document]
ruivieira: AFAIK it's not. I agree with thinking of a better way of doing it.

danielezonca (approved):

PR #24, opened 2023-06-06
Use ConfigMap to customise TrustyAI operator's service image

ruivieira:
[RHODS-9182](https://issues.redhat.com/browse/RHODS-9182) / #23 

Configure the operator to use custom images by creating a `ConfigMap` in the operator's namespace. 
The operator only checks the `ConfigMap` at deployment, so changes made afterward won't trigger a redeployment of services.

Example ConfigMap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: trustyai-service-operator-config
data:
  trustyaiServiceImageName: 'quay.io/mycustomrepo/mycustomimage'
  trustyaiServiceImageTag: 'v1.0.0'
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -99,21 +99,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Get image and tag from ConfigMap
-	// If there's a ConfigMap with custom images, it is only applied when the operator is first deployed
-	// Changing (or creating) the ConfigMap after the operator is deployed will not have any effect
-	imageName, imageTag, err := controllers.GetConfigMapValues(mgr.GetClient(), ns, "trustyai-service-operator-config")
-	if err != nil {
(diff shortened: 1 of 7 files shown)


tteofili (approved):
LGTM

main.go
```diff
@@ -93,9 +93,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	ns, err := controllers.GetNamespace()
+	if err != nil {
+		setupLog.Error(err, "unable to operator's namespace")
+		os.Exit(1)
```
danielezonca, line 99: Please don't make this a final error: this feature is mainly for testing so I would prefer to log (Error is fine or Warn as you prefer) with proper message but default on the native scenario (aka hardcoded image)
Wdyt?

ruivieira: Changed in latest commit

PR #29, opened 2023-06-07
Ensure PV exists and create PVC with matching storage class

ruivieira:
[RHODS-9247](https://issues.redhat.com/browse/RHODS-9247) / #26 

The operator now:

- Ensures the specified PV exists (and logs approriate errors)
- Uses the same storage class as the PV for the PVC to avoid any errors


tteofili (approved):
LGTM

danielezonca (approved):

PR #42, opened 2023-06-08
Add TLS and route checking

ruivieira:


controllers/route.go
```diff
@@ -0,0 +1,89 @@
...
+		} else {
+			// They are different, update the existing route.
+			existingRoute.Spec = createdRoute.Spec
+			err := r.Client.Update(ctx, existingRoute)
+			if err != nil {
+				log.FromContext(ctx).Error(err, "Error updating Route.")
+				return err
+			}
+		}
+	}
```
danielezonca, line 86: No `Condition` to represent `Route` status? This should impact also on general readiness of the service.
Feel free to address with a future ticket :)

danielezonca (approved):
LGMT, just a comment but that can be addressed with another ticket

controllers/route.go
ruivieira: 👍 https://github.com/trustyai-explainability/trustyai-service-operator/issues/43

PR #55, opened 2023-06-19
Update ModelMesh deployments with TrustyAI payload processor endpoints

ruivieira:


controllers/trustyaiservice_controller.go
```diff
@@ -138,10 +143,15 @@ func (r *TrustyAIServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	// Call the function to patch environment variables for Deployments that match the label
+	shouldContinue, err := r.patchEnvVarsByLabelForDeployments(ctx, req.Namespace, modelMeshLabelKey, modelMeshLabelValue, payloadProcessorName, req.Name, false)
 	if err != nil {
-		log.FromContext(ctx).Error(err, "Failed to update ModelMesh payload processor.")
-		// handle error
+		log.FromContext(ctx).Error(err, "Could not patch environment variables for Deployments.")
+		return ctrl.Result{}, err
+	}
+	if !shouldContinue {
+		log.FromContext(ctx).Info("Not all replicas are ready, requeue the reconcile request")
```
danielezonca, line 153: This is more a side comment but why do we need to wait all replicas to be up before doing the injection of the variable? Is there any undesired side effect or it _just_ don't work? :)

controllers/modelmesh.go
```diff
@@ -2,98 +2,114 @@ package controllers
...
+	if err != nil {
+		log.FromContext(ctx).Error(err, "Could not get Deployments by label.")
+		return false, err
+	}
+
+	// Build the payload processor endpoint
+	url := "http://" + crName + "." + namespace + ".svc.cluster.local/consumer/kserve/v2"
+
+	// Patch environment variables for the Deployments
+	if shouldContinue, err := r.patchEnvVarsForDeployments(ctx, deployments, envVarName, url, remove); err != nil {
```
danielezonca, line 76: Minor note, I think this code is good but the `remove` codepath is a bit hard to understand: the logic is hidden inside `patchEnvVarsForDeployments` so for example the `url` is built even while removing.
I think it is fine for now, we can probably improve it in the future but I would like to minimise the changes at this stage :)

controllers/modelmesh.go
```diff
@@ -2,98 +2,114 @@ package controllers
...
+		log.FromContext(ctx).Error(err, "Could not patch environment variables for Deployments.")
+		return shouldContinue, err
+	}
+
+	return true, nil
+}
+
+// updateEnvVarValue updates the value of an environment variable based on the remove flag and current value
+func updateEnvVarValue(currentValue, newValue string, remove bool) string {
```
danielezonca, line 85: I think "update" is a bit misleading, because as far as I see the method calculate the new value but it is not doing the actual update.
Am I right?

controllers/modelmesh.go
```diff
@@ -2,98 +2,114 @@ package controllers
...
-					newParts := []string{}
-					for _, part := range parts {
-						if part != url {
-							newParts = append(newParts, part)
-						}
-					}
-					envVar.Value = strings.Join(newParts, " ")
-				}
+				// Modify the existing env var based on the remove flag and current value
+				envVar.Value = updateEnvVarValue(envVar.Value, url, remove)
```
danielezonca, line 49: Where this value is updated? Is it line 53 correct?

controllers/modelmesh.go
ruivieira: Renamed to `generateEnvVarValue` and updated the docs

controllers/modelmesh.go
ruivieira: https://github.com/trustyai-explainability/trustyai-service-operator/issues/56

controllers/finalizers.go
```diff
@@ -4,11 +4,5 @@ import "context"
...
 // deleteExternalDependency removes the payload processor from the ModelMesh deployment
 func (r *TrustyAIServiceReconciler) deleteExternalDependency(crName, namespace string, ctx context.Context) error {
-	// Remove the payload processor from the deployment
-	err := updatePayloadProcessor(context.Background(), r.Client, "mlserver", "PAYLOAD_PROCESSOR", crName, namespace, true)
-	if err != nil {
-		return err
-	}
-
 	return nil
 }
```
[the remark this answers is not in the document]
ruivieira: Removed.

controllers/modelmesh.go
ruivieira: Yes, the last step of the loop.

controllers/finalizers.go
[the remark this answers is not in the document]
ruivieira: Reinstated finaliser

controllers/trustyaiservice_controller.go
```diff
@@ -577,6 +587,23 @@ func (r *TrustyAIServiceReconciler) reconcileServiceMonitor(cr *trustyaiopendata
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *TrustyAIServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &appsv1.Deployment{}, ".metadata.controller", func(rawObj client.Object) []string {
```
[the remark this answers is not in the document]
ruivieira: Changed to `context.Background()`

controllers/trustyaiservice_controller.go
ruivieira: It will work, this is the previous logic (no longer needed). It will be removed with
- #57

PR #68, opened 2023-06-29
Change import paths and API references

ruivieira:


go.sum
danielezonca, line 1: are you sure about this changes? Given that you changed only some import I was not expecting impact on this 🤔

danielezonca (approved):

go.sum
ruivieira: Reverted the changes to `go.sum`.

PR #72, opened 2023-07-05
Correct finalizer name

ruivieira:


controllers/trustyaiservice_controller.go
```diff
@@ -44,7 +44,7 @@ const (
 	defaultTag           = string("latest")
 	containerName        = "trustyai-service"
 	serviceMonitorName   = "trustyai-metrics"
-	finalizerName        = "trustyai.opendatahub.io/finalizer"
+	finalizerName        = "trustyai.opendatahub.io.trustyai.opendatahub.io/finalizer"
```
danielezonca, line 47: Is it possible to use a shared constant to avoid/prevent similar issues?

danielezonca (approved):

controllers/trustyaiservice_controller.go
ruivieira: - https://github.com/trustyai-explainability/trustyai-service-operator/issues/73

PR #81, opened 2023-07-14
Add central ServiceMonitor to operator's namespace

ruivieira:
This PR creates an additional service monitor on the operator's namespace.
It will watch namespace for `app.kubernetes.io/part-of: trustyai-service` CRs and configure the metrics.
The previous per-namespace Service Monitor is still created (as previously).


controllers/monitor.go
```diff
@@ -0,0 +1,169 @@
...
+
+// generateServiceMonitorSpecCentral generates the ServiceMonitor spec for central ServiceMonitor
+func generateServiceMonitorSpecCentral(deploymentNamespace string) *monitoringv1.ServiceMonitor {
+	serviceMonitor := &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceMonitorName,
+			Namespace: deploymentNamespace,
+			Labels: map[string]string{
+				"modelmesh-service": "modelmesh-serving",
+			},
```
danielezonca, line 24: Is this label used by some part of serving monitoring stack to filter?

controllers/monitor.go
```diff
@@ -0,0 +1,169 @@
...
+	// Create only one central ServiceMonitor
+	if isCentralServiceMonitorCreated {
+		return nil
+	}
+
+	serviceMonitor := generateServiceMonitorSpecCentral(r.Namespace)
+
+	// Check if this ServiceMonitor already exists
+	found := &monitoringv1.ServiceMonitor{}
+	err := r.Get(ctx, types.NamespacedName{Name: serviceMonitor.Name, Namespace: serviceMonitor.Namespace}, found)
```
danielezonca, line 72: Can you please clarify this part? As far as I see the `isCentralServiceMonitorCreated` is just a shortcut to avoid to check if the resource already exists.
Am I right?
I'm not sure if it is reliable: for example the controller can be restarted (so the variable set to `false` again) or even worst the `ServiceMonitor` CR can be manually cancelled and it should be recreated.

If it is only an optimization I would prefer to stay on the safe side and do the `GET` every time for now.
Wdyt?

controllers/monitor.go
```diff
@@ -0,0 +1,169 @@
...
+
+// generateServiceMonitorSpecLocal generates the ServiceMonitor spec for a local ServiceMonitor
+func generateServiceMonitorSpecLocal(deploymentNamespace string, serviceName string) *monitoringv1.ServiceMonitor {
+	serviceMonitor := &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceName,
+			Namespace: deploymentNamespace,
+			Labels: map[string]string{
+				"modelmesh-service": "modelmesh-serving",
+			},
```
danielezonca, line 104: Similar comment about label here

controllers/monitor.go
```diff
@@ -0,0 +1,169 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var isCentralServiceMonitorCreated = false
+
+// generateServiceMonitorSpecCentral generates the ServiceMonitor spec for central ServiceMonitor
+func generateServiceMonitorSpecCentral(deploymentNamespace string) *monitoringv1.ServiceMonitor {
```
danielezonca, line 17: As far as I see the owner ref is not configured for this resource.
Given that it is created in the same namespace of the controller I would set the controller as owner.
Wdyt?

controllers/monitor.go
```diff
@@ -0,0 +1,169 @@
...
+
+	serviceMonitor := generateServiceMonitorSpecCentral(r.Namespace)
+
+	// Check if this ServiceMonitor already exists
+	found := &monitoringv1.ServiceMonitor{}
+	err := r.Get(ctx, types.NamespacedName{Name: serviceMonitor.Name, Namespace: serviceMonitor.Namespace}, found)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.FromContext(ctx).Info("Creating a new central ServiceMonitor", "ServiceMonitor.Namespace", serviceMonitor.Namespace, "ServiceMonitor.Name", serviceMonitor.Name)
+			err = r.Create(ctx, serviceMonitor)
```
danielezonca, line 76: Should we also configure a watcher for this resource to trigger reconcile?

controllers/monitor.go
ruivieira: Good point, perhaps it's not needed.
Moved to 
- https://github.com/trustyai-explainability/trustyai-service-operator/issues/82

controllers/monitor.go
ruivieira: Removed the boolean flag. Central SM is not part of the reconcile loop.

controllers/monitor.go
ruivieira: - https://github.com/trustyai-explainability/trustyai-service-operator/issues/83

controllers/monitor.go
danielezonca: The reason of my comment is mainly to solve the situation of someone deleting central ServiceMonitor. It should never happen but we can consider to have a watcher to trigger reconcile and recreate it in this case.
Btw up to you :)

danielezonca (approved):

PR #91, opened 2023-07-21
Refactor reconcile return values

ruivieira:


controllers/trustyaiservice_controller.go
```diff
@@ -97,10 +97,10 @@ func (r *TrustyAIServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			return ctrl.Result{}, nil
+			return Requeue()
```
danielezonca, line 100: This seems a different behaviour

ruivieira: @danielezonca well spotted, changed to not requeueing, since this is the scenario where it might have been deleted.

danielezonca (approved):

PR #105, opened 2023-09-25
Add KServe support

ruivieira:
This PR adds support to KServe inference services (IS) by setting a TrustyAI service as an `InferenceLogger` for KServe IS.
If IS is part of a ModelMesh deployment, the behaviour is the same as previously.

KServe and ModelMesh IS are differentiated by checking the IS's `serving.kserve.io/deploymentMode: ModelMesh` annotation.

**Other changes**

- Dependencies updated to better align with the ODH operator
- Updated KServe API from `v1alpha1` to `v1beta1`


config/rbac/role.yaml
```diff
@@ -107,6 +107,27 @@ rules:
...
+  resources:
+  - inferenceservices
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
danielezonca, line 121: Are really all of them necessary? For example I don't think `create` or `delete` are necessary

danielezonca (approved):

config/rbac/role.yaml
ruivieira: Removed

PR #156, opened 2023-11-17
Add Available condition and checks

ruivieira:
This PR introduces the `Available` condition.
This condition is set to `True` if the PVC, Route and Deployment are ready.
The definition of ready is:
- PVC bound
- Route is admitted
- Deployment is Available and with expected number of replicas
The `Available` condition would be `False` otherwise.

`InferenceServices` do not have a bearing in `Available` and whether they are present or not will be reflected in the `InferenceServicesPresent` condition.

The readiness status checks are moved to the `reconcileStatuses` method and performed together, rather than at each component's (PVC, Route, etc) check.


ruivieira:
cc @alexcreasy

tteofili (approved):

PR #185, opened 2024-01-24
[RHOAIENG-2135] Add unique ClusterRoleBinding names

ruivieira:
Fixes [RHOAIENG-2135](https://issues.redhat.com/browse/RHOAIENG-2135)

### Problem

When creating multiple `TrustyAIServices`, endpoint authentication will fail.

### Steps to reproduce

Steps to reproduce using the CLI:

Assuming:
- A TrustyAI operator already deployed
- `$CLUSTER` is your cluster address.
- A `$TOKEN` as `export TOKEN=$(oc whoami -t)`
- No models are required to reproduce this

Create a namespace `test`. 
Add a first `TrustyAIService` in this namespace , using, for instance (`trustyai-test.yaml`)

```
apiVersion: trustyai.opendatahub.io/v1alpha1
kind: TrustyAIService
metadata:
  name: trustyai-service
spec:
  storage:
    format: "PVC"
    folder: "/inputs"
    size: "1Gi"
  data:
    filename: "data.csv"
    format: "CSV"
  metrics:
    schedule: "5s"
```
 and

```
oc apply -f trustyai-test.yaml -n test
```

Try the external endpoint:

```
curl -sk -H "Authorization: Bearer ${TOKEN}" https://trustyai-service-test.${CLUSTER}/info
```

This should provide a successful response (in terms of authentication, actual response might be `[]`).

Create a namespace `test2`. 
Add a `TrustyAIService` in this namespace , using, for instance the `trustyai-test.yaml` as before:

```
oc apply -f trustyai-test.yaml -n test2
```

Try the external endpoint:

```
curl -sk -H "Authorization: Bearer ${TOKEN}" https://trustyai-service-test2.${CLUSTER}/info
```

The request will now be unsuccessful in terms of authentication.

### Solution

For authentication, TrustyAI's proxy uses a `ServiceAccount` with a binding to a proxy role with `tokenreviews` and `subjectaccessreviews`. This binding was created for each service and with a name derived from the TrustyAI service name.
If more than one service with the same were deployed, this would create a conflict and cause the second role binding to not be created.

`ClusterRoleBinding` names are now unique, derived from the service's name and namespace.

### Validation

Re-run the entire "Steps to reproduce" section.
The final request should now be successful

```
curl -sk -H "Authorization: Bearer ${TOKEN}" https://trustyai-service-test2.${CLUSTER}/info
```

PR #196, opened 2024-02-06
RHOAIENG-2417: Remove unnecessary cluster role binding's namespace

ruivieira:
See https://github.com/trustyai-explainability/trustyai-service-operator/pull/185#discussion_r1464612407


ruivieira:
This needs to be further tested.

PR #237, opened 2024-04-22
RHOAIENG-6342: Add ODH and RHOAI overlays

ruivieira:
[RHOAIENG-6342](https://issues.redhat.com/browse/RHOAIENG-6342).

Add overlays specifically for ODH and RHOAI.

### DIFF
--- a/config/overlays/odh/kustomization.yaml
+++ b/config/overlays/odh/kustomization.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base
+configMapGenerator:
+  - env: params.env
+    behavior: merge
+    name: config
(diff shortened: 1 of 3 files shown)


danielezonca (approved):

zdtsw:
i dont see problem with this PR,  but wonder params.env will work for both ODH and RHOAI (for the future).
maybe just keep params.env in the base, 
and keep the overlay/odh and overlay/rhoai.
in the ODH case: base + overlay/odh
for the RHOAI case: base + overlay/rhoai.
you can keep updating base/params.env with your quay.io image for ODH
and we will take care of the internal build SHA1 in base/params.env for RHOAI.

ruivieira:
> i dont see problem with this PR, but wonder params.env will work for both ODH and RHOAI (for the future). maybe just keep params.env in the base, and keep the overlay/odh and overlay/rhoai. in the ODH case: base + overlay/odh for the RHOAI case: base + overlay/rhoai. you can keep updating base/params.env with your quay.io image for ODH and we will take care of the internal build SHA1 in base/params.env for RHOAI.

@zdtsw thanks! The purpose of RHOAI is not to change the images themselves, but to add some RHOAI-specific configuration to the deployments in the future.

zdtsw (approved):

PR #272, opened 2024-08-04
feat: Add LMES driver build to GHA

ruivieira:
Add LMES drive image build to GHA.


yhwang (approved):
thanks for the change
/LGTM

PR #280, opened 2024-08-21
feat: Driver updates job's status periodically

yhwang:
The driver periodically update the LMEvalJob.Status.Message field with the outputs from the lm-eval. The message pattern the driver captures is like `Running text generation:  81%|`. Then users can use this information to check the progress of the job.

### DIFF
--- a/Dockerfile.driver
+++ b/Dockerfile.driver
@@ -1,7 +1,16 @@
 FROM registry.access.redhat.com/ubi8/go-toolset:1.21 AS builder
 
 WORKDIR /go/src/github.com/trustyai-explainability/trustyai-service-operator
-COPY . .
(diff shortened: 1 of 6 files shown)


yhwang:
/retest

to see if it's an intermittent failure

ruivieira:
@yhwang It seems the error happens due to https://github.com/trustyai-explainability/trustyai-service-operator/pull/280/files#diff-ff3da6967ee893290caa6ed4cf523cca9bf4109109e969e4c8422a4bc22054e2R114 redefining `zap-devel`.
Is it possible to make the devel option for the logger global to all test, for instance?

yhwang:
@ruivieira nice catch! fixed it. Thanks!

PR #283, opened 2024-08-28
feat: Add overlays

ruivieira:
Add overlays to support selective controller configuration:

- default, `odh` and `lmes-tas` overlays enable both `TAS` and `LMES`
- `rhoai` overlay enables `TAS` only (disables `LMES`)


config/base/params.env
```diff
@@ -1,4 +1,10 @@
 trustyaiServiceImage=quay.io/trustyai/trustyai-service:latest
 trustyaiOperatorImage=quay.io/trustyai/trustyai-service-operator:latest
 oauthProxyImage=quay.io/openshift/origin-oauth-proxy:4.14.0
-kServeServerless=disabled
\ No newline at end of file
+kServeServerless=disabled
+lmes-driver-image=quay.io/trustyai/ta-lmes-driver:latest
+lmes-pod-image=quay.io/trustyai/lm-eval-aas-flask:latest
```
yhwang, line 6: I believe you will update this to `ta-lmes-job`.

config/overlays/lmes-tas/params.env
```diff
@@ -1,6 +1,3 @@
-trustyaiServiceImage=quay.io/trustyai/trustyai-service:v0.13.0
-trustyaiOperatorImage=quay.io/trustyai/trustyai-service-operator:v1.19.0
-oauthProxyImage=quay.io/openshift/origin-oauth-proxy:4.14.0
 lmes-driver-image=quay.io/yhwang/ta-lmes-driver:latest
 lmes-pod-image=quay.io/yhwang/lm-eval-aas-flask:test
```
yhwang, line 2: same `ta-lmes-job` here.

yhwang (approved):
Nice overlays!
Just one minor question:
In your change, the base is `TAS+LMES` and ODH inherits the base. RHOAI overrides it to `TAS` only.
The other way around is to keep the base as `TAS` only and change ODH to `TAS+LMES`, then for RHOAI, it can inherit from the base directly.

These two approaches all deliver the same outcome. I guess the key point is:
the base should be `TAS` only or `TAS + LMES`.  This may be also related to how the e2e test deploys the operator. I will leave this to you. I am good with either way.

config/overlays/lmes-tas/params.env
yhwang: it would be good to update both `lmes-driver-image` and `lmes-pod-image` here

config/base/params.env
ruivieira: That's right, thanks. This was writen before we decided on the image's name 👍

ruivieira:
> Nice overlays! Just one minor question: In your change, the base is `TAS+LMES` and ODH inherits the base. RHOAI overrides it to `TAS` only. The other way around is to keep the base as `TAS` only and change ODH to `TAS+LMES`, then for RHOAI, it can inherit from the base directly.
> 
> These two approaches all deliver the same outcome. I guess the key point is: the base should be `TAS` only or `TAS + LMES`. This may be also related to how the e2e test deploys the operator. I will leave this to you. I am good with either way.

Thanks @yhwang that's a good point.

Personally, I think we should keep the base as LMES+TAS. ODH can just inherit from base.
We then provide LMES only for specific use cases. Any downstream can then configure the controllers in their respective forks.
In this case, the overlay for LMES+TAS is in fact redundant.

I will change the GHA to build the `ta-lmes-job` in a separate PR.

zdtsw:
i am good with this PR, looks neat and only small changes required from Platform when enable this feature.

Issue #288, opened 2024-09-05
Allow users to specify the batch size for the LMEvalJob


yhwang:
lm-evaluation-harness provides the `--batch-size` option to specify the batch size for the evaluation which can speed up the evaluation process. It would be good to support this option in the LMEvalJob CR.

yhwang:
In the LMEvalJob CR, a new `batchSize` field can be added for this purpose. For the lm-eval controller, it would be good to add a `max-batch-size` config and use it as the cap for the `batchSize` that users can specify.

yhwang:
Close this because the PR above has been merged.

PR #290, opened 2024-09-09
feat: support batch size

yhwang:
fix: #288
Add batch size support in the LMEvalJob which
leverages the `--batch_size` in the `lm-evaluation-harness`. This only affects the local models. The `--bath_size` doesn't work for remote inference APIs.


yhwang:
add a new line at the end of some customization YAML files to fix the YAML linter errors.

ruivieira (approved):
LGTM, thanks!

PR #296, opened 2024-09-13
fix: fix dependency error in the job image

yhwang:
Split up the unitxt and openai dependencies to avoid the conflict.


yhwang:
one thing to add is that: manually installing the `unitxt` would be removed once we upgrade to lm-eval v0.4.4, since it was added to the core dependeicnes. We will only need to install the `openai` by then.

ruivieira (approved):
LGTM

PR #309, opened 2024-10-04
feat: support custom dataset

yhwang:
Updated the CRD data struct to allow users to specify a custom Unitxt card in JSON format. The custom Unitxt card is equivalent to a custom dataset definition. Also restructured and updated the CRD to support Volumes, VolumeMounts, Env, Resources, Labels, and Annotations.


tedhtchang:
Looks good. I may soon need the containers resources request object to create a PodSet (for the Kueue integration).

tedhtchang (approved):

PR #313, opened 2024-10-09
Add initial Kueue integration

tedhtchang:
This PR is to optionally enable Kueue for LMEvalJob by implementing the Kueue's GenericJob interface. It still needs some other functionalities so I want to broke it down a little bit:
### 1. Kueue requires LMEvalJob  to have Spec.Suspend. We need to handle these 3 suspend cases in the LMEvalJob reconciler: 
   1. Spec.Suspend(=true) upon LMEvalJob creation. This means the job is created but pods are running.
   2. Spec.Suspend(=true) while the LMEvalJob is running. This means delete the pods owned by the job.
   3. Spec.Suspend(=false) while the LMEvalJob is in suspended state. This means re-create the pods owned by the job.
See [PR#317](https://github.com/trustyai-explainability/trustyai-service-operator/pull/317)

### 2. Get the PodSpec from the LMEValJob.  Currently PodSpec is not stored in the LMEvalJob object. 
Kueue needs to admit the job as a [kueue.x-k8s.io/v1beta1/Workload](https://kueue.sigs.k8s.io/docs/concepts/workload/). So the Workload can represent the resources (cpu/memory) requirements for the job.
We need the  PodSpec from the LMEvalJob object to create a PodSet for Workload. See [PR#323](https://github.com/trustyai-explainability/trustyai-service-operator/pull/323)

### 3. Node Affinity
When we want to run job on particular node. We can create a spec.nodeLabels in ResourceFlavor which can be referenced in a ClusterQueue. 
```yaml
apiVersion: kueue.x-k8s.io/v1beta1
kind: ResourceFlavor
metadata:
  name: "default-flavor"
spec:
  nodeLabels:
    kubernetes.io/hostname: kueue-worker
```
Implement these [runWithPodInfo and restoreWithPodInfo](https://github.com/trustyai-explainability/trustyai-service-operator/blob/6639b9be1d4ed82368bab767c647713b24ec551e/controllers/job_mgr/job_mgr_controller.go#L86-L97).
Extract nodeSelector object from from the PodSetInfo object and then overwrite the LMevalJob PodSpec spec affinity. @yhwang  created  another [PR](https://github.com/trustyai-explainability/trustyai-service-operator/pull/340)  to introduce [Pod.Spec.NodeAffinity ](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#nodeaffinity-v1-core) into LMevalJob.


ruivieira:
Hi @tedhtchang this is looking great, thanks!

Just a question, we cannot guarantee that Kueue is installed on the cluster. Would enabling Kueue support for LM-Eval (and not having Kueue installed) be a problem?

tedhtchang:
@ruivieira @yhwang Can you guys change the PR to merge to main branch Or I can close & open a new PR?

tedhtchang:
> Hi @tedhtchang this is looking great, thanks!
> 
> Just a question, we cannot guarantee that Kueue is installed on the cluster. Would enabling Kueue support for LM-Eval (and not having Kueue installed) be a problem?

@ruivieira We can start the operator without the job_mgr_controller if Kueue is not enabled. It shouldn't be a problem.

cmd/main.go
```diff
@@ -69,7 +72,7 @@ func main() {
 	var probeAddr string
 	var configMap string
 	var enabledServices controllers.EnabledServices
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":9443", "The address and port the metric endpoint binds to.")
```
yhwang, line 75: Switching the default value here to 9443 may conflict with the webhook port although we don't have a webhook server yet. Actually, the metric port is not exposed either.

controllers/job_mgr/README.md
```diff
@@ -0,0 +1,235 @@
...
+        internal.config.kubernetes.io/generatorBehavior: unspecified
+        internal.config.kubernetes.io/prefixes: trustyai-service-operator-
+        internal.config.kubernetes.io/previousKinds: ConfigMap,ConfigMap
+        internal.config.kubernetes.io/previousNames: config,trustyai-service-operator-config
+        internal.config.kubernetes.io/previousNamespaces: default,default
+    data:
+        kServeServerless: disabled
+        lmes-default-batch-size: "8"
+        lmes-driver-image: quay.io/yhwang/ta-lmes-driver:latest
+        lmes-grpc-port: "8082"
```
yhwang, line 121: nit: no more gRPC settings

controllers/job_mgr/constants.go
```diff
@@ -0,0 +1,5 @@
+package job_mgr
```
yhwang, line 1: missing the license header.

controllers/job_mgr/job_mgr_controller.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (job *LMEvalJob) Suspend() {
+	job.Spec.Suspend = true
+}
+
+// RunWithPodSetsInfo will inject the node affinity and podSet counts extracting from workload to job and unsuspend it.
+func (job *LMEvalJob) RunWithPodSetsInfo(podSetsInfo []podset.PodSetInfo) error {
+	job.Spec.Pod.Affinity = convertToAffinity(podSetsInfo)
+	job.Spec.Suspend = false
```
yhwang, line 82: one question: so this function needs you to update the `job.Spec.Suspend` to `false` but no need to set the `job.Spec.Suspend` to `true` in the `RestorePodSetsInfo` function?

controllers/lmes/constants.go
```diff
@@ -40,4 +40,5 @@ const (
 	DefaultBatchSize           = 8
 	DefaultDetectDevice        = true
 	ServiceName                = "LMES"
+	EnableKueue                = false
```
yhwang, line 43: I don't see where this is used.

controllers/lmes/config.go
```diff
@@ -27,7 +27,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-var options *serviceOptions = &serviceOptions{
+// set by job_mgr controllerSetup func
+var Job_mgr_enabled bool
```
yhwang, line 31: nit: change to CamelCase

controllers/job_mgr/job_mgr_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return
+}
+
+// PodSets will build workload podSets corresponding to the job.
+func (job *LMEvalJob) PodSets() []kueue.PodSet {
+	log := log.FromContext(context.TODO())
+	pod := lmes.CreatePod(lmes.Options, &job.LMEvalJob, log)
+	podSet := kueue.PodSet{
+		Name:     job.Status.PodName,
```
yhwang, line 119: `job.Status.PodName` is updated after the LMES controller creates the job pod. Is this function triggered after the job pod is created?

controllers/job_mgr/job_mgr_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+// GVK returns GVK (Group Version Kind) for the job.
+func (job *LMEvalJob) GVK() schema.GroupVersionKind {
+	return workloadv1alpha1.GroupVersion.WithKind("LMEvalJob")
+}
+
+func convertToAffinity(psi []podset.PodSetInfo) *corev1.Affinity {
+	// Note there is only 1 element in podset array see PodSets method above.
```
yhwang, line 152: one question: no possibility that the length of the `psi` is 0?

yhwang (commented):
Great efforts! Left some comments and questions. After this, we need to discuss how to handle the new controller `JOB_MGR` in the overlays.

controllers/job_mgr/job_mgr_controller.go
tedhtchang: I believe PodSet is created(for Workloads) before the Pod can created.

controllers/job_mgr/job_mgr_controller.go
tedhtchang: I wondered that too. I am not sure what triggers this RestorePodSetsInfo function. I will add it for now.

controllers/job_mgr/job_mgr_controller.go
```diff
@@ -0,0 +1,177 @@
...
+			nsr := corev1.NodeSelectorRequirement{
+				Key:      k,
+				Operator: "In",
+				Values:   []string{v},
+			}
+			nsra = append(nsra, nsr)
+		}
+		nsta := []corev1.NodeSelectorTerm{}
+		nsta = append(nsta, corev1.NodeSelectorTerm{MatchExpressions: nsra})
+		affinity := corev1.Affinity{
```
yhwang, line 167: nit: you can return here
```suggestion
		return &corev1.Affinity{
```
and remove the return at #174

yhwang, line 167:
```suggestion
		return &corev1.Affinity{
```
nit: you can return here
and remove the return at #174

controllers/job_mgr/job_mgr_controller.go
tedhtchang: I observed RestorePodSetsInfo is called right before a job preempted But if I had job.Spec.Suspend = false the job would not be evicted. So I removed it again.

api/lmes/v1alpha1/lmevaljob_types.go
```diff
@@ -282,6 +315,11 @@ type LMEvalJob struct {
 	Status LMEvalJobStatus `json:"status,omitempty"`
 }
 
+// generate pod name for the job
+func (j *LMEvalJob) GetPodName() string {
```
yhwang, line 319: nit: since you have this function now, can you also update lmevaljob_controller.go#411 to use this function?

tedhtchang: done

controllers/job_mgr/job_mgr_controller.go
tedhtchang: done

controllers/lmes/lmevaljob_controller.go
```diff
@@ -168,7 +184,7 @@ func (r *LMEvalJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		job.Status.State = lmesv1alpha1.NewJobState
 	}
 
-	if job.Spec.Suspend {
+	if job.Spec.Suspend && JobMgrEnabled {
```
yhwang, line 187: What should we do if `JobMgrEnabled` is `true` but there is no `kueue.x-k8s.io/queue-name` label in the LMEvalJob? Does 'no Kueue label' mean using a regular approach to create a pod without Kueue's resource management?

tedhtchang: Done. Per discussion no label should mark job complete with failed reason and messages.

controllers/lmes/lmevaljob_controller.go
```diff
@@ -180,14 +180,22 @@ func (r *LMEvalJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// Treat this as NewJobState
-	if job.Status.LastScheduleTime == nil {
+	if job.Status.LastScheduleTime == nil && job.Status.CompleteTime == nil {
```
yhwang, line 183: it's great to update the condition here, nice!

controllers/lmes/lmevaljob_controller.go
```diff
@@ -510,18 +518,20 @@ func (r *LMEvalJobReconciler) deleteJobPod(ctx context.Context, job *lmesv1alpha
...
 	if job.Status.CompleteTime == nil {
 		// make sure the pod is in the complete state. if not, run the shutdown command
-		pod, err := r.getPod(ctx, job)
-		if err == nil {
-			if getRunningContainerByName(&pod.Status, "main") != -1 {
-				// send shutdown command if the main container is running
-				if err := r.shutdownDriver(ctx, job); err != nil {
-					log.Error(err, "failed to shutdown the job pod. retry after the pulling interval")
-					return r.pullingJobs.addOrUpdate(string(job.GetUID()), Options.PodCheckingInterval), nil
+		if job.Status.PodName != "" {
```
yhwang, line 521: with this condition check, the handleComplete function can properly handle the case where the job pod hasn't been created yet. nice

yhwang (approved):
/LGTM

Great work! With this change, LMES can use Kueue for job management (optional).

tedhtchang:
> Hi @tedhtchang , is this PR rebased on `main`? If not, is it ok to rebase it? Thank you!

@ruivieira You should merge the [PR](https://github.com/trustyai-explainability/trustyai-service-operator/pull/340) first because my PR was based on it. After you merged it, I will rebased the main into my PR. Then you should be able to merged mine.

ruivieira:
@tedhtchang ack, thanks for the heads up. I've approved https://github.com/trustyai-explainability/trustyai-service-operator/pull/340 now.

ruivieira (approved):

tedhtchang:
@ruivieira @yhwang I rebased and resolved conflicts from main and update-lmevaljob-pod branches.

controllers/tas/statuses_test.go
```diff
@@ -51,7 +52,10 @@ func setupAndTestStatusNoComponent(instance *trustyaiopendatahubiov1alpha1.Trust
 var _ = Describe("Status and condition tests", func() {
 
 	BeforeEach(func() {
-		k8sClient = fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+		k8sClient = fake.NewClientBuilder().WithScheme(scheme.Scheme).WithStatusSubresource(
+			&routev1.Route{},
+			&trustyaiopendatahubiov1alpha1.TrustyAIService{},
```
tedhtchang, line 57: Need to register the status subresoruce separately with the updated controller-runtime package.
Thanks @yhwang figured this out.

tedhtchang:
@ruivieira The controller-test failure was fix. Not sure about the smoke test error. Does it  always fail ?

ruivieira (changes requested):
@tedhtchang LGTM, thanks for the changes!

I do have a final request: is it possible to regenerate the `go.sum`?
I'm getting a checksum mismatch for `github.com/kserve/kserve@v0.12.1` dependency.

tedhtchang:
@ruivieira I saw no difference in go.sum  after `go mod tidy`. I rebase PR anyway.
however @yhwang found mismatched go.sum in the main branch. This is the output:
```
go mod tidy
go: downloading golang.org/x/mod v0.12.0
go: downloading github.com/grpc-ecosystem/grpc-gateway/v2 v2.15.2
```

PR #314, opened 2024-10-10
feat: new pulling mechanism for job statuses

yhwang:
Update the driver to keep running even the user program finishes. The driver provides two APIs:
- GetStatus(): retrieve job status
- Shutdown(): properly tear down the driver

In the controller side, it uses `pod/exec` resource to run the driver command to invoke the driver APIs to retrieve the job status and shutdown the driver when job is done.


yhwang:
Some info about the pulling mechanism:
The controller uses `pod/exec` to run the driver command on the job pod to retrieve the job status and results periodically (using the `lmes-pod-checking-interval` setting). The driver command uses the domain socket as the IPC to exchange the information with the driver that runs the evaluation job on the pod. When a job is done, the controller uses the same way to run the driver command on the pod to shut down the running driver.
The communication between the controller and job pods goes through the Kube API server. No more gRPC and ingress traffic on the operator.

tedhtchang (approved):
/LGTM I tested with my suspend PR
```
ENABLED_SERVICES=LMES CONTROLLER_TOOLS_VERSION=v0.16.3 IMG=quay.io/tedchang/trustyai-service-operator:latest make run
```
In another terminal:
```
cat << EOF| kubectl apply -f -
kind: LMEvalJob
metadata:
  labels:
    app.kubernetes.io/name: fms-lm-eval-service
    app.kubernetes.io/managed-by: kustomize
  name: evaljob-sample
  namespace: default
spec:
  suspend: false
  model: hf
  modelArgs:
  - name: pretrained
    value: EleutherAI/pythia-70m
  taskList:
    taskNames:
    - unfair_tos
  logSamples: true
  limit: "5"
EOF
```
Verified job and pod in running state

ruivieira (approved):
LGTM, thanks @yhwang!

PR #318, opened 2024-10-13
Change job's user ID to fix SCC violation

ruivieira:
Refers to [RHOAIENG-13846](https://issues.redhat.com/browse/RHOAIENG-13846).

The pod is trying to run with a runAsUser UID of 1001030000, which is outside the allowed UID ranges specified by the Security Context Constraints available to the pod's service account.

This PR sets it to a UID within the allowed range `[1000840000, 1000849999]`.

### DIFF
--- a/controllers/lmes/lmevaljob_controller.go
+++ b/controllers/lmes/lmevaljob_controller.go
@@ -76,6 +76,11 @@ var (
 	annotationFilterPrefixes = []string{}
 )
 
+// Constants
+const (
+	jobUserId int64 = 1000840000
+)
+
 // maintain a list of key-time pair data.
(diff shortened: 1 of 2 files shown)


RobGeada (approved):

ruivieira:
@yhwang I've removed the user id in https://github.com/trustyai-explainability/trustyai-service-operator/pull/322 and in my testing it works fine. Letting OpenShift automatically assign the user id (based on SCC and the relevant namespace range) seems to be enough.

We can move any discussion to https://github.com/trustyai-explainability/trustyai-service-operator/pull/322, but in the meantime I'm closing this PR.

PR #323, opened 2024-10-15
Refactor some lmesreconcile methods

tedhtchang:
Refactors createPod other related methods to static functions because I need to create PodSet for Kueue integration using those functions


yhwang (commented):
thanks for fixing typos and refactoring the `serviceOptions` as a singleton. good change

yhwang (approved):
/LGTM

PR #332, opened 2024-10-18
Support for VirtualServices for InferenceLogger traffic

ruivieira:
* Create `VirtualService`s automatically for Inference Logger traffic
* Rename Inference Logger endpoint to conform with destination rules and virtual services


controllers/inference_services.go
```diff
@@ -295,6 +295,25 @@ func (r *TrustyAIServiceReconciler) patchKServe(ctx context.Context, instance *t
 			log.FromContext(ctx).Error(err, "InferenceService has service mesh annotation but DestinationRule CRD not found")
 		}
 
+		// Check if VirtualService CRD is present. If there's an error, don't proceed and return the error
```
danielezonca, line 298: We should double check if the creation of the VirtualService is not producing any side effect like exposing the endpoint outside the namespace or similar.
I think we can do that with a follow-up

danielezonca (approved):
Two comments but approving because they can be done in the future

controllers/inference_services.go
ruivieira: Thanks @danielezonca, from my POV this isn't the case, as we discussed elsewhere.
Still I'm noting this for a more comprehensive review.

PR #350, opened 2024-10-30
RHOAIENG-14773: Change batchSize to string in order to support "auto" and "auto:N"

ruivieira:
Refers to [RHOAIENG-14773](https://issues.redhat.com/browse/RHOAIENG-14773).

This PR changes the field `batchSize` from `int` to `string` in order to support the values `auto` and `auto:N`.
Validation is performed by the controller for the following scenarios:

* `batchSize="n"`, assuming `n` is a positive integer it will be used, otherwise use the default batch size
* `batchSize="auto"`, `auto` will be used
* `batchSize="auto:n"`, if `n` is a positive integer `auto:n` will be used and the maximum batch size will be ignored (`n` will be the maximum batch size). If `n` is invalid (negative, or not a `int`), `auto:maxBatchSize` will be used.
* `batchSize="n"`, (where `n > maxBatchSize`) maximum batch size will be used

This PR also lower the default batch size from `8` to `1`.

### DIFF
--- a/controllers/lmes/lmevaljob_controller.go
+++ b/controllers/lmes/lmevaljob_controller.go
@@ -845,7 +845,7 @@ func validateBatchSize(input string, maxBatchSize int, log logr.Logger) string {
 		// If N is valid, but larger than maxBatchSize, set it to maximum batch size
 		if n > maxBatchSize {
 			log.Info("batchSize is greater than max-batch-size of the controller's configuration, use the max-batch-size instead")
-			return strconv.Itoa(maxBatchSize)
+			return maxBatchSizeString
 		}
 		// If N is valid, use it
 		return strconv.Itoa(n)
--- a/controllers/lmes/lmevaljob_controller.go
+++ b/controllers/lmes/lmevaljob_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


controllers/lmes/lmevaljob_controller.go
```diff
@@ -818,6 +819,42 @@ func mergeMapWithFilters(dest, src map[string]string, prefixFilters []string, lo
...
+		log.Info(input + " not supported. Using auto:" + maxBatchSizeString)
+		return "auto:" + maxBatchSizeString
+	}
+
+	// Validate N batch size
+	if n, err := strconv.Atoi(input); err == nil && n > 0 {
+		// If N is valid, but larger than maxBatchSize, set it to maximum batch size
+		if n > maxBatchSize {
+			log.Info("batchSize is greater than max-batch-size of the controller's configuration, use the max-batch-size instead")
+			return strconv.Itoa(maxBatchSize)
```
yhwang, line 848: nit: isn't this the `maxBatchSizeString`?

controllers/lmes/lmevaljob_controller.go
```diff
@@ -818,6 +819,42 @@ func mergeMapWithFilters(dest, src map[string]string, prefixFilters []string, lo
...
+		}
+		// If N is valid, use it
+		return strconv.Itoa(n)
```
yhwang, line 851: I assume this will be the `input` but not sure if there is any circumstance that `Atoi` -> `Itoa` will modify the positive int string. but I agree your approach is safer.

controllers/lmes/lmevaljob_controller_test.go
```diff
@@ -1365,3 +1365,30 @@ func Test_PVCPreference(t *testing.T) {
 
 	assert.Equal(t, expect, newPod)
 }
+
+func Test_ValidateBatchSize(t *testing.T) {
```
yhwang, line 1369: nice test case!

yhwang (approved):
/LGTM

left some minor comments and overall it's good for me.

controllers/lmes/lmevaljob_controller.go
ruivieira: Well spotted, removed the redundant conversion.

yhwang (approved):

PR #351, opened 2024-11-04
RHOAIENG-15105: Add offline support for LMEvalJobs

ruivieira:
Refers to [RHOAIENG-15105](https://issues.redhat.com/browse/RHOAIENG-15105).

Add support for offline flags in the `LMEvalJob` container and mount PVC containing models and datasets.

If the optional `offline` field is set:

```
 offline:
    storage:
      pvcName: pvc-with-my-models
```

offline mode will be enabled. If it is missing, offline mode will be disable.


yhwang (approved):
/LGTM

While reviewing the code, it reminded me of one thing not directly related to this change. I believe it would be good to create an issue regarding enabling the validation webhook for LMES controller. Then we can enable some tedious value checking there.

ruivieira:
@yhwang Opened https://github.com/trustyai-explainability/trustyai-service-operator/issues/353

PR #355, opened 2024-11-07
fix(RHOAIENG-14321): Revert metrics service target port to https

ruivieira:
Reverts the operator's metrics service target port to `https`.


nehachopra27 (approved):
LGTM

PR #356, opened 2024-11-08
fix(test): Allow smoke test to pass

tedhtchang:
Allow smoke test to pass


ruivieira (approved):
Well spotted! LGTM, thanks @tedhtchang

PR #358, opened 2024-11-11
fix(lmeval): Send Job logs to Pod StdOut

RobGeada:


controllers/lmes/driver/driver.go
```diff
@@ -283,6 +283,29 @@ func (dc *driverComm) notifyShutdownWait() {
...
 }
 
+type PodLogAndPipeWriter struct {
+	pw *io.PipeWriter
+}
+
+func (f PodLogAndPipeWriter) Write(p []byte) (n int, err error) {
+	// format lm-eval output to stdout and the pipewriter
+	line := string(p)
+	res, _ := fmt.Print(line)
```
yhwang, line 293: since this writer prints out the stderr, maybe line 134 should be commented out. or the stderr would be printed again when the lm-eval job finishes.

controllers/lmes/driver/driver.go
```diff
@@ -283,6 +283,29 @@ func (dc *driverComm) notifyShutdownWait() {
...
+	if _, err := f.pw.Write(p); err != nil {
+		return 0, err
+	}
+
+	// carriage returns do not correctly display, so replace with newlines
+	// carriage returns also break the pipewriter, so make sure we get newlines as well.
+	if strings.ContainsAny(line, "\r") && !strings.ContainsAny(line, "\n") {
```
yhwang, line 300: I am thinking would it be more straightforward to directly do a `regexp.ReplaceAll()` for all incoming []byte i.e:
```
re := regexp.MustCompile(`(?m)\r[^\n]`)
updatedBytes := re.ReplaceAll(p, []byte("\n"))
```

yhwang (approved):
Thanks for fixing this Rob. So the problem is the output pipe. I thought it was the scanner.

controllers/lmes/driver/driver.go
ruivieira: @yhwang I've opened https://github.com/trustyai-explainability/trustyai-service-operator/issues/359 so this could be done on a separate PR after this one is tested, if that's ok.

ruivieira (approved):

controllers/lmes/driver/driver.go
yhwang: sounds good to me

PR #363, opened 2024-11-14
feat(lmeval): Enable Kueue Job Manager in LMEval

ruivieira:
Enable LMEval's Job Manager for Kueue support

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -33,7 +33,7 @@ spec:
           args:
             - --leader-elect
             - --enable-services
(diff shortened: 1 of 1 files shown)


yhwang (approved):
/LGTM

yhwang:
The only concern is that this will implicitly make TrustyAI-service-operator depend on Kueue. I guess this may need more discussion.

ruivieira:
Thanks @yhwang, this is an important point.

I think it's best to close this PR and, for the moment, add the job manager support as an optional overlay on a separate PR.

PR #364, opened 2024-11-14
feat(lmeval): Add overlay for Kueue support

ruivieira:
Add `odh-kueue` overlay with job manager enabled for Kueue support.


yhwang (approved):
/LGTM

It's nice to have the overlay for JOB_MGR.

PR #409, opened 2025-02-12
feat(LMES): Switch LMES driver to a container port

yhwang:
Originally, the LMES driver used a Unix domain socket to serve the APIs which are used by the LMES controller to retrieve the job status. Switch to a container port to allow other containers in the same job pod to access the APIs as well.


ruivieira (approved):
LGTM, thanks @yhwang

PR #435, opened 2025-03-27
feat(RHOAIENG-22430): Add custom YAML task support to LMEval

ruivieira:
This PR adds support for external/custom lm-evaluation-harness tasks.
Only git repositories are accepted as a remote location, for now.

To define an external tasks we can use a CR such as (example)

```yaml
kind: LMEvalJob
metadata:
  name: lmeval-llama-stack-job
  namespace: test
spec:
  allowCodeExecution: true
  allowOnline: true
  batchSize: "1"
  logSamples: true
  model: hf
  modelArgs:
  - name: pretrained
    value: "google/flan-t5-base"
  - name: tokenizer
    value: "google/flan-t5-base"
  taskList:
    customTasks:
      source:
        git:
          url: https://github.com/trustyai-explainability/lm-eval-tasks.git
          branch: main
          commit: "b995eeb986daab3c9f77f101af6cc51a34f8f1da"
          path: tasks/
    taskNames:
      - tiny_offline_arithmetic
      - arc_easy
```

The field introduced with this PR is `customTasks`. It provides the following configurations:

- `source`: the location of the external/custom tasks to be added (only `git` supported for now)
- `git`: specifies a git storage
- `url`: the git storage URL
- `branch`: branch where the tasks are located (defaults to `main`)
- `commit`: a specific commit to checkout (defaults to `HEAD`)
- `path`: path within the repository where tasks are located

The new tasks can be specified along "builtin" ones in `taskNames`, the example above shows a mixture of external and builtin (arc_easy)

In the above example, we are specifying that the tasks are present at https://github.com/trustyai-explainability/lm-eval-tasks git URL, on the `main` branch at commit `b995eeb986daab3c9f77f101af6cc51a34f8f1da`, under the path `tasks`.
We also specify that we want to run the `tiny_offline_arithmetic` evaluation.

The LMEval driver will then clone this repository, moving the tasks to the already existing include path.

This PR also respects the `allowOnline: true/false` flag, meaning that if online is disable, the remote tasks will not be fetched, resulting in an error.


config/crd/bases/trustyai.opendatahub.io_lmevaljobs.yaml
```diff
@@ -4757,6 +4757,38 @@ spec:
...
+                                type: string
+                              path:
+                                description: Path specifies the path to the task file
+                                type: string
+                              url:
+                                description: URL specifies the git repository URL
+                                type: string
+                            type: object
+                        type: object
+                      taskNames:
```
RobGeada, line 4785: Can we merge this with the parent taskNames field? This would mirror the lm-eval-harness api:

Using a default task:
```
lmeval/bin/lm_eval \
        --tasks $default-task-name \
```

Using a custom task:
```
lmeval/bin/lm_eval \
        --tasks $custom-task-name \
        --include_path $custom-task-directory \
```
So `taskNames` can be names from either the default set or the custom task sources

RobGeada (changes requested):

config/crd/bases/trustyai.opendatahub.io_lmevaljobs.yaml
yhwang: I am also thinking where the `customTasks` should be added. Here is my thought:

Currently we have the following fields under `taskList`:
```go
type TaskList struct {
	// TaskNames from lm-eval's task list
	TaskNames []string `json:"taskNames,omitempty"`
	// Task Recipes specifically for Unitxt
	TaskRecipes []TaskRecipe `json:"taskRecipes,omitempty"`
	// Custom Unitxt artifacts that can be used in a TaskRecipe
	CustomArtifacts *CustomArtifacts `json:"custom,omitempty"`
}
```
Under `custom`, there are custom templates, custom system_prompts, and custom metrics(coming soon). However, these custom artifacts are using the `CustomArtifact` data type, and you can only assign a string value as the custom value. So should we:
- Expand the CustomArtifact data type to also support `GitSource` and add the `customTasks` under the `custom` and rename it to `tasks`. Or just use `[]CustomTaskSource` as its data type and add it to `CustomArtifacts`?
- Or we differentiate the external source as `external` and put it next to `custom`?

I feel the custom tasks in this PR are different from the custom templates, custom system_prompts, and metrics. But somehow adding GitSource to CustomArtifact seems also a good feature.

yhwang: BTW, I also like @RobGeada 's idea. Because `customTasks` are not like custom templates, custom system_prompts, and custom metrics, they are only used under `taskList`. Updating the taskList.taskNames` from []string to a new type which contains []string and GitSource is also a good idea.

ruivieira: @RobGeada @yhwang for now I've updated the PR to support a single `spec.taskList.taskNames` (i.e. we declare "external" task locations in `customTasks`, but then we treat it as any builtin task).

@yhwang That's an interesting idea. However, IIUC there is a conceptual difference between `CustomArtifact` (I'm defining something on the CR) and `CustomTask` (I'm pointing to a location, no definition). Perhaps the `CustomTask` could be renamed to make this clearer?

yhwang: Agree. like I mentioned earlier, I do feel `CustomTask` and `CustomArtifacts` are different. Maybe change to `externalTask`?

The other reason I mentioned `CustomArtifacts` here is that I do think `CustomTaskSource` could be generalized to `ExternalSource` and be used by `CustomArtifacts` as well.

cmd/lmes_driver/main.go
```diff
@@ -54,13 +54,19 @@ var (
...
 	customSystemPrompts strArrayArg
+	taskNames           strArrayArg
 	copy                = flag.String("copy", "", "copy this binary to specified destination path")
 	getStatus           = flag.Bool("get-status", false, "Get current status")
 	shutdown            = flag.Bool("shutdown", false, "Shutdown the driver")
 	outputPath          = flag.String("output-path", OutputPath, "output path")
 	detectDevice        = flag.Bool("detect-device", false, "detect available device(s), CUDA or CPU")
 	commPort            = flag.Int("listen-port", driver.DefaultPort, "driver serves APIs on the port")
 	downloadAssetsS3    = flag.Bool("download-assets-s3", false, "Download assets from S3")
+	customTaskGitURL    = flag.String("custom-task-git-url", "", "Git repository URL for custom tasks")
```
yhwang, line 65: I am thinking of merging several args into one, i.e., `externalSource`, and its value could be a JSON object that we can marshal/unmarshal using a struct type directly or a format you see fit. To generalize it, one extra info is needed: where these external sources are copied to. And it could accept multiple values, which means you can specify this argument multiple times. With this new argument, the driver can support multiple external resource retrievals.

RobGeada (approved):

PR #438, opened 2025-04-11
RHOAIENG-23547: Decouple vllm-gateway and regex detector

christinaexyou:
This PR addresses [RHOAIENG-23547](https://issues.redhat.com/browse/RHOAIENG-23547) by making the following changes:

* Adds two new boolean flags to the GORCH CR to enable the regex detectors and vLLM gateway sidecars separately: `regexDetectors` and `vllmGateway`
* Adds the regex detectors and vLLM gateway image tags to `params.env`
* Reads the image references to both from the `trustyai-service-config` configmap
* Updates tests to reflect changes

### DIFF
--- a/api/gorch/v1alpha1/guardrailsorchestrator_types.go
+++ b/api/gorch/v1alpha1/guardrailsorchestrator_types.go
@@ -34,9 +34,15 @@ type GuardrailsOrchestratorSpec struct {
 	Replicas int32 `json:"replicas"`
 	// Name of configmap containing generator,detector,and chunker arguments
 	OrchestratorConfig *string `json:"orchestratorConfig"`
-	//  Name of the configmap containg vLLM gateway arguments
+	// Boolean flag to enable/disable built-in detectors
 	// +optional
-	VLLMGatewayConfig *string `json:"vllmGatewayConfig,omitempty"`
+	EnableBuiltInDetectors bool `json:"enableBuiltInDetectors,omitempty"`
+	// Boolean flag to enable/disable the guardrails sidecar gateway
+	// +optional
+	EnableGuardrailsGateway bool `json:"enableGuardrailsGateway,omitempty"`
+	//  Name of the configmap containing guadrails sidecar gateway arguments
+	// +optional
+	SidecarGatewayConfig *string `json:"guardrailsGatewayConfig,omitempty"`
 	// List of orchestrator enviroment variables for configuring the OTLP exporter
 	// +optional
 	OtelExporter OtelExporter `json:"otelExporter,omitempty"`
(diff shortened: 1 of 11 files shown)


api/gorch/v1alpha1/guardrailsorchestrator_types.go
```diff
@@ -34,6 +34,12 @@ type GuardrailsOrchestratorSpec struct {
 	Replicas int32 `json:"replicas"`
 	// Name of configmap containing generator,detector,and chunker arguments
 	OrchestratorConfig *string `json:"orchestratorConfig"`
+	// Boolean flag to enable/disable the regex detectors
+	// +optional
+	RegexDetectors bool `json:"regexDetectors,omitempty"`
```
ruivieira, line 39: Would it be possible to group this into a key called (say) `detectors`?
If more detectors will be provided OOTB, the spec will get quite cluttered in the future.

config/base/params.env
```diff
@@ -12,3 +12,5 @@ lmes-detect-device=true
 lmes-allow-online=true
 lmes-allow-code-execution=true
 guardrails-orchestrator-image=quay.io/trustyai/ta-guardrails-orchestrator:latest
+guardrails-regex-detector-image=quay.io/trustyai/regex-detector:latest
```
ruivieira, line 15: Should we reverse the naming? i.e. `guardrails-detector-${detector_name}-image`.
It will be simpler for grouping in the future.

ruivieira (changes requested):
Thanks @christinaexyou , LGTM, left a couple of suggestions

api/gorch/v1alpha1/guardrailsorchestrator_types.go
christinaexyou: Renamed to `EnableDetectors` and removed the prefix "regex" for any variables related to detectors

config/base/params.env
christinaexyou: Renamed to `guardrails-detector-image`

ruivieira (approved):

controllers/gorch/constants.go
```diff
@@ -1,11 +1,11 @@
...
-	orchestratorImageKey  = "guardrails-orchestrator-image"
-	vllmGatewayImageKey   = "vllmGatewayImage"
-	regexDetectorImageKey = "regexDetectorImage"
-	ServiceName           = "GORCH"
+	orchestratorName     = "guardrails-orchestrator"
+	finalizerName        = "trustyai.opendatahub.io/gorch-finalizer"
+	configMapName        = "gorch-config"
+	orchestratorImageKey = "guardrails-orchestrator-image"
+	gatewayImageKey      = "guardrails-sidecar-gateway-image"
+	detectorImageKey     = "guardrails-builtin-detector-image"
```
ruivieira, line 9: Typo: should be `guardrails-built-in-detector-image`

ruivieira (changes requested):

ruivieira (approved):

Issue #446, opened 2025-04-16
Support RAG metric with retrieval system


yhwang:
Unitxt already supports RAG metrics. However, the way it integrates with LMEH is assuming the retrieval information is put into the dataset with the proper column names, i.e. "contexts", "answer". So the evaluation process gets the pre-retrieved information from the dataset and run the evaluation procedures as usual and calculate the metrics.

To support an end-to-end RAG scenario, it needs a mechanism to integrate with the retrieval system. Each retrieval system behaves differently, and LMEvalJob is not able to support all variations. Fortunately, MCP (model context protocol) is becoming more and more popular because more and more external systems are going to integrate with LLM nowadays. Here is the potential approach:
- Extend the current Unitxt task implementation in the LMEH to support information retrieval using MCP
- Merge retrieved information with the dataset
- Run the evaluation as usual.

The benefits are:
- Setting up an MCP server to support the existing knowledge database is fairly easy.
- Instead of support different kind knowledge dbs, we leverage MCP as a standard way to support various retrieval systems.
- No need to prepare an extra dataset that contains pre-retrieved columns from the retrieval system.

PR #511, opened 2025-06-24
feat: support RAG end-2-end scenario

yhwang:
Add RAG end-to-end support by leveraging the following features:
- Use the RAG support from the unitxt
- Extend the unitxt RAG to call out the MCP tool for information retrieval

related: #446


yhwang:
This change requires an additional change in the unitxt to support the MCP tool calling as the information retrieval mechanism for the RAG scenario. Here is the [unitxt PR](https://github.com/kiersten-stokes/lm-evaluation-harness/pull/2) for that. I believe using the MCP tool for information retrieval is a good approach to support the end-to-end RAG scenario. If we have an agreement on this, we can integrate the unitxt enhancement into the downstream lm-evaluation-harness repo. Then this PR would fulfill the controller change and leverage the new feature.

yhwang:
The improvement/change in the MCP SDK moves swiftly, especially in the security aspect, because there was a lack of security at the beginning. The implementation here aims to support `HTTPS` at least, and it would be good to support OAuth if possible. I labeled this as `Draft` and would like to get feedback about the mechanism established in this PR. Then I can work on it to finalize the implementation.

PR #519, opened 2025-07-09
Update params.env

ruivieira:


tarilabs (approved):
bumps and set Env for the operator image variable to 1.40.0 following https://github.com/trustyai-explainability/trustyai-service-operator/pull/519

/lgtm
/approve

PR #525, opened 2025-07-16
fix: Check pod readiness before accepting completion status from the driver

ruivieira:
* Added a check to verify that the pod is ready before accepting a `Complete` status from the driver
* Added logging to clarify when we status updates are ignored

## Summary by Sourcery

Verify pod readiness before accepting completion status from the driver and add logging for ignored status updates

Bug Fixes:
- Prevent premature completion by checking that the pod’s main container is running before accepting a Complete status
- Log informative messages when completion updates are ignored because the pod is still initializing


sheltoncyril (approved):
/lgtm

PR #530, opened 2025-07-30
feat: Add subject to LMEval CRD's OCI outputs

ruivieira:
## Summary by Sourcery

Add support for an OCI subject in LMEvalJob outputs by extending the OCISpec type, updating the CRD schema, injecting the OCI_SUBJECT env var in pods, and adding a corresponding unit test

New Features:
- Introduce Subject field in OCISpec to allow specifying OCI artifact subjects
- Expose OCI_SUBJECT environment variable in pods when a Subject is provided

Documentation:
- Update CRD schema and documentation to include the optional Subject property

Tests:
- Add unit test to verify OCI_SUBJECT environment variable is correctly set

### DIFF
--- a/api/lmes/v1alpha1/lmevaljob_types.go
+++ b/api/lmes/v1alpha1/lmevaljob_types.go
@@ -481,6 +481,7 @@ type OCISpec struct {
 	// Subject for the OCI artifact
 	// +optional
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9._:/@-]*$`
+	// +kubebuilder:validation:MaxLength=255
 	Subject string `json:"subject,omitempty"`
 	// Username for registry authentication
 	// +optional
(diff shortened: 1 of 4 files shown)


ruivieira:
@sourcery-ai review

api/lmes/v1alpha1/lmevaljob_types.go
```diff
@@ -478,6 +478,11 @@ type OCISpec struct {
 	// Path within the results to package as artifact
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9._/-]*$`
 	Path string `json:"path"`
+	// Subject for the OCI artifact
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9._:/@-]*$`
+	// +kubebuilder:validation:MaxLength=255
+	Subject string `json:"subject,omitempty"`
```
tarilabs, line 485: as this will be a OCI reference, I would validate that this reference starts with `registry` as prefix, wdyt?

ie, if
```
OCI_REGISTRY=quay.io
OCI_REPOSITORY=mmortari/my-artifact
```

then the
```
OCI_SUBJECT=quay.io/mmortari/my-model
```

starts with the same "server" OCI Registry (`quay.io`).

note: I don't find in the spec that the "org" must be the same, but it's likely a convention implemented by many registry, which I would not validate and eventually we response-code-fail if the permission setup from the server rejects it. I will followup on this point tho.

PR #543, opened 2025-09-04
fix(RHOAIENG-33645): Pass --confirm-run-unsafe-code arg when AllowCodeExecution is enabled

christinaexyou:
This PR updates the LMEval controller and tests to pass in `--confirm-run-unsafe-code` as an argument when `AllowCodeExecution` is set to `true`. Additionally, to allow code execution during evaluation, we also pass in an env var `HF_ALLOW_CODE_EVAL` set to 1.

## Summary by Sourcery

Enable unsafe code execution in LMEvalJobs by adding the HF_ALLOW_CODE_EVAL environment variable and passing the --confirm_run_unsafe_code flag when the AllowCodeExecution option is enabled

New Features:
- Pass --confirm_run_unsafe_code argument to LMEval controller when AllowCodeExecution is true

Enhancements:
- Inject HF_ALLOW_CODE_EVAL environment variable into LMEval pods and add it to the protected env var list

Tests:
- Update existing tests to assert HF_ALLOW_CODE_EVAL values in different modes
- Add Test_AllowCodeExecution to verify the confirmation flag is included

### DIFF
--- a/controllers/lmes/lmevaljob_controller.go
+++ b/controllers/lmes/lmevaljob_controller.go
@@ -928,6 +928,10 @@ func CreatePod(svcOpts *serviceOptions, job *lmesv1alpha1.LMEvalJob, log logr.Lo
 			Name:  "UNITXT_ALLOW_UNVERIFIED_CODE",
 			Value: "False",
 		},
+		{
+			Name:  "HF_ALLOW_CODE_EVAL",
+			Value: "0",
+		},
 	}
 	allowRemoteCodeEnvVars := []corev1.EnvVar{
 		{
@@ -941,6 +945,10 @@ func CreatePod(svcOpts *serviceOptions, job *lmesv1alpha1.LMEvalJob, log logr.Lo
 		{
 			Name:  "UNITXT_ALLOW_UNVERIFIED_CODE",
 			Value: "True",
+		},
+		{
+			Name:  "HF_ALLOW_CODE_EVAL",
+			Value: "1",
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/lmes/lmevaljob_controller.go
```diff
@@ -942,6 +942,10 @@ func CreatePod(svcOpts *serviceOptions, job *lmesv1alpha1.LMEvalJob, log logr.Lo
 			Name:  "UNITXT_ALLOW_UNVERIFIED_CODE",
 			Value: "True",
 		},
+		{
+			Name:  "HF_ALLOW_CODE_EVAL",
+			Value: "1",
+		},
```
ruivieira, line 948: If this is dependent on the allowRemote field, should we explicitly add

```go
	{
			Name:  "HF_ALLOW_CODE_EVAL",
			Value: "0",
		},
```

to `disallowRemoteCodeEnvVars`?

controllers/lmes/lmevaljob_controller.go
```diff
@@ -1361,6 +1365,11 @@ func generateArgs(svcOpts *serviceOptions, job *lmesv1alpha1.LMEvalJob, log logr
 		}
 	}
 
+	// --confirm_run_unsafe_code
+	if job.Spec.AllowCodeExecution != nil && *job.Spec.AllowCodeExecution {
+		cmds = append(cmds, "--confirm_run_unsafe_code")
```
ruivieira, line 1370: Should we add explicit logging here? We can add to other flags too, in retrospect, but in another PR.

```go
log.Info("Enabling unsafe code execution for LMEval job", "jobName", job.Name,  "namespace", job.Namespace)
```

ruivieira (changes requested):
LGTM @christinaexyou, thanks!

Just a couple of comments.

PR #561, opened 2025-09-18
fix(RHOAIENG-30427): Default LMEval permission read from default config, when DSC config is missing

ruivieira:
Refer to [RHOAIENG-34517](https://issues.redhat.com/browse/RHOAIENG-34517).

### DIFF
--- a/controllers/dsc/config.go
+++ b/controllers/dsc/config.go
@@ -50,7 +50,7 @@ func (r *DSCConfigReader) ReadDSCConfig(ctx context.Context, log *logr.Logger) (
 		if errors.IsNotFound(err) {
 			log.V(1).Info("DSC ConfigMap not found, using default configuration",
 				"configmap", configMapKey)
-			return &DSCConfig{}, nil
+			return nil, nil
 		}
 		return nil, fmt.Errorf("error reading DSC ConfigMap %s: %w", configMapKey, err)
 	}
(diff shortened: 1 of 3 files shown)


sheltoncyril (approved):

sheltoncyril (changes requested):
Think this does not work, just tested it with devFlags

controllers/dsc/config.go
```diff
@@ -50,7 +50,7 @@ func (r *DSCConfigReader) ReadDSCConfig(ctx context.Context, log *logr.Logger) (
 		if errors.IsNotFound(err) {
 			log.V(1).Info("DSC ConfigMap not found, using default configuration",
 				"configmap", configMapKey)
-			return &DSCConfig{}, nil
```
tarilabs, line 53: shouldn't here return an error as well? 🤔 otherwise later in:

https://github.com/trustyai-explainability/trustyai-service-operator/blob/6b4663033aa7269fc07a3ea1f5c06881ec4d2211/controllers/lmes/lmevaljob_controller.go#L280-L287

in case this CM is not found, you end up again in line 286, meaning you apply `false`, `false` on top of the trustyai-service-operator-config values?

tarilabs (commented):
thanks Rui,

controllers/dsc/config.go
tarilabs: nevermind I retract that, I see:

https://github.com/trustyai-explainability/trustyai-service-operator/blob/6b4663033aa7269fc07a3ea1f5c06881ec4d2211/controllers/lmes/config.go#L120-L126

PR #588, opened 2025-10-15
Feat: Add servicemonitor for built-in detectors

RobGeada:
## Summary by Sourcery

Implement ServiceMonitor support for built-in detectors by creating ServiceMonitor resources in the reconciler when enabled and conditionally exposing a detector-metrics port; standardize component labels across templates for consistent resource selection.

New Features:
- Add ServiceMonitor creation logic in GuardrailsOrchestratorReconciler when EnableBuiltInDetectors is true
- Introduce a service-monitor template and createServiceMonitor helper function
- Conditionally expose a detector-metrics port on the service spec based on the built-in detectors flag

Enhancements:
- Standardize component labels to "trustyai-guardrails" across service, deployment, gateway, health, https-route, and CA bundle templates

### DIFF
--- a/controllers/gorch/constants.go
+++ b/controllers/gorch/constants.go
@@ -4,7 +4,6 @@ const (
 	orchestratorName     = "guardrails-orchestrator"
 	finalizerName        = "trustyai.opendatahub.io/gorch-finalizer"
 	configMapName        = "gorch-config"
-	serviceMonitorName   = "trustyai-guardrails-monitor"
 	orchestratorImageKey = "guardrails-orchestrator-image"
 	gatewayImageKey      = "guardrails-sidecar-gateway-image"
 	detectorImageKey     = "guardrails-built-in-detector-image"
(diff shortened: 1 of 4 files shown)


controllers/gorch/constants.go
```diff
@@ -4,6 +4,7 @@ const (
 	orchestratorName     = "guardrails-orchestrator"
 	finalizerName        = "trustyai.opendatahub.io/gorch-finalizer"
 	configMapName        = "gorch-config"
+	serviceMonitorName   = "trustyai-guardrails-monitor"
```
m-misiura: is there a mismatch between constant and template name? I think the template generates `{{ .Orchestrator.Name }}-service-monitor`?

controllers/gorch/templates/service-monitor.tmpl.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
```
m-misiura: should `namespace` be included in the metadata section?

PR #634, opened 2026-02-05
Add admin, edit, and view role aggregation to both guardrails controllers

RobGeada:
## Summary by Sourcery

Add aggregated RBAC roles for guardrails orchestrators and integrate them into the operator's RBAC configuration.

New Features:
- Introduce editor and viewer ClusterRoles for guardrailsorchestrators resources with appropriate CRUD and read-only permissions.

Enhancements:
- Enable aggregation of nemo guardrail editor and viewer roles into standard Kubernetes admin, edit, and view aggregate roles.
- Include new guardrails orchestrator editor and viewer roles in the RBAC kustomization configuration.

## Summary by CodeRabbit

* **Chores**
  * Added cluster-level RBAC roles (admin, editor, viewer) for Guardrails Orchestrator to standardize access.
  * Introduced cluster-level admin, editor, and viewer roles for NeMo-Guardrails and replaced previous NeMo-Guardrails role manifests with consolidated cluster-level roles.
  * Removed legacy NeMo-Guardrails editor and viewer role manifests.


m-misiura (approved):

danielezonca (changes requested):
Please don't merge yet.
I would like @lburgazzoli or @lphiri to review

ugiordan (changes requested):
Currently, users with `admin` or `edit` roles in their namespace can't create GuardrailsOrchestrator (RBAC blocks it), but after this change, the aggregation labels would allow them to create it in their namespace.

Is this the intended behavior? Just want to confirm that namespace-level management of these resources is the goal.

RobGeada:
Hi @ugiordan @danielezonca - sorry for the delay. The intention is for the permissions for the guardrail components to exactly match permissions for models- if you have permissions to create a model deployment, you should have permissions to create a guardrails deployment, etc.

To this end, I've updated the PR so that the RBAC and aggregation for both guardrails components align with [KServe's RBAC](https://github.com/opendatahub-io/kserve/blob/master/config/overlays/odh/rbac/user-cluster-roles.yaml).

m-misiura (approved):

config/components/gorch/rbac/guardrailsorchestrator_user_cluster_roles.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - apiGroups:
+      - trustyai.opendatahub.io
+    resources:
+      - guardrailsorchestrators
+      - guardrailsorchestrators/status
+      - guardrailsorchestrators/finalizers
+    verbs:
+      - get
+      - list
+      - watch
```
ugiordan, line 37: This editor role only has `get/list/watch`, making it identical to the viewer below. Missing: create, delete, patch, update.
The nemoguardrails-editor-role in this same PR correctly has all 7 verbs, so this looks like a regression from the latest force-push. Without the write verbs, namespace editors still can't create/modify `GuardrailsOrchestrator` resources, which defeats the purpose of the PR.

config/components/gorch/rbac/guardrailsorchestrator_user_cluster_roles.yaml
```diff
@@ -0,0 +1,57 @@
...
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-guardrailsorchestrator-admin-role: "true"
+  name: guardrailsorchestrator-editor-role
+rules:
+  - apiGroups:
+      - trustyai.opendatahub.io
+    resources:
+      - guardrailsorchestrators
+      - guardrailsorchestrators/status
+      - guardrailsorchestrators/finalizers
```
ugiordan: The editor role should only have CRUD on the main resource. Including `/status` and `/finalizers` here means editors get write access to controller-managed state.
In the [KServe reference](https://github.com/opendatahub-io/kserve/blob/master/config/overlays/odh/rbac/user-cluster-roles.yaml) you're following, the edit role only operates on main resources. `/status` and `/finalizers` appear only in the viewer (read-only). Our existing `trustyaiservice-editor-role` also restricts `/status` to get-only.

We should split into two rules, or remove `/status` and `/finalizers` from the editor entirely.

config/components/gorch/rbac/guardrailsorchestrator_user_cluster_roles.yaml
```diff
@@ -0,0 +1,57 @@
...
+---
+# permissions for end users to view FMS Guardrails Orchestrators.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: trustyai-service-operator
+    app.kubernetes.io/managed-by: kustomize
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-guardrailsorchestrator-admin-role: "true"
```
ugiordan: In the KServe pattern, only the edit role aggregates into the custom admin shell. The viewer should only aggregate into the standard view (no `aggregate-to-guardrailsorchestrator-admin-role` label here). This label makes the viewer role changes silently propagate to admins.

config/components/nemo-guardrails/rbac/nemoguardrails_user_cluster_roles.yaml
```diff
@@ -0,0 +1,61 @@
...
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-nemoguardrails-admin-role: "true"
+  name: nemoguardrails-editor-role
+rules:
+  - apiGroups:
+      - trustyai.opendatahub.io
+    resources:
+      - nemoguardrails
+      - nemoguardrails/status
+      - nemoguardrails/finalizers
```
ugiordan: Same as the gorch editor: `/status` and `/finalizers` shouldn't be in the editor role with full CRUD verbs. KServe's edit only operates on main resources. The old `nemoguardrail-editor-role` being replaced here correctly restricted `/status` to get-only.

We should remove `/status` and `/finalizers` from this rule, or split into separate rules where `/status` gets get-only access.

config/components/nemo-guardrails/rbac/nemoguardrails_user_cluster_roles.yaml
```diff
@@ -0,0 +1,61 @@
...
+---
+# permissions for end users to view NeMo-Guardrails.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: trustyai-service-operator
+    app.kubernetes.io/managed-by: kustomize
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-nemoguardrails-admin-role: "true"
```
ugiordan: Same as the gorch viewer: this `aggregate-to-nemoguardrails-admin-role` label should be removed. KServe's view role does not aggregate into the custom admin.

config/components/gorch/rbac/guardrailsorchestrator_user_cluster_roles.yaml
RobGeada: yep, that's copy-paste mistake, fixed

config/components/gorch/rbac/guardrailsorchestrator_user_cluster_roles.yaml
RobGeada: Got it, will fix

config/components/nemo-guardrails/kustomization.yaml
```diff
@@ -6,5 +6,4 @@ resources:
   - crd/trustyai.opendatahub.io_nemoguardrails.yaml
   - rbac/manager-rbac.yaml
   - rbac/manager-rolebinding.yaml
-  - rbac/nemoguardrail_editor_role.yaml
-  - rbac/nemoguardrail_viewer_role.yaml
+  - rbac/nemoguardrails_user_cluster_roles.yaml
```
ugiordan, line 9: The rename from `nemoguardrail-editor-role` / `nemoguardrail-viewer-role` to `nemoguardrails-editor-role` / `nemoguardrails-viewer-role` will orphan the old ClusterRoles on existing clusters (kustomize tracks by name). Worth noting in the PR description or documenting a cleanup step.

RobGeada: good point - we might as well just keep the names the same?

ugiordan: The ClusterRole names inside the file are already singular (`nemoguardrail-editor-role`, `nemoguardrail-viewer-role`), which is what matters for avoiding orphaned resources on the cluster. The filename itself doesn't affect Kubernetes, so either naming works. Up to you.

RobGeada: yep - I just updated them to singular, they weren't when you made the comment

PR #704, opened 2026-04-06
Detect kueue workload failures and update eval job as failed

nbs-rh:
When the eval-hub API caller submits a job with a queue name that does not exist on the cluster, a kube job is created in 'suspended' state and a workload CR is created (same as in case of a valid queue name being passed). However, in case of queue name being invalid, the error message is only updated in the `conditions` section of the workload, so the eval job remained in `pending` state forever. This PR has the changes to detect such workloads in evalhub controller and update the job as 'failed' with the appropriate reason.

Related to https://redhat.atlassian.net/browse/RHOAIENG-53404

Testing:
#### A job was submitted with invalid queue name. `GET job` API returned failed status with the error message.

image

#### On a cluster without kueue CRD installed, the workload reconciler is not registered/does not start. 

```
nbs@nbs-mac ~ % oc logs -f trustyai-service-operator-controller-manager-5f698cd7cc-jnvvn -n opendatahub | grep "Kueue Workload API not available" 
2026-04-07T03:38:11Z	INFO	Kueue Workload API not available; skipping EvalHub failed Kueue Workloads controller	{"groupVersion": "kueue.x-k8s.io/v1beta1", "resource": "workloads"}
```

If a job is submitted with a queue name (valid or not), eval-hub service still sets the kueue label but the job is still processed by the default scheduler:

image

## Summary by CodeRabbit

* **New Features**
  * Detect and report Kueue workload failures to EvalHub via a new reconciler; improved end‑to‑end failure reporting and retries.
  * Shared, concurrency‑safe tenant namespace cache for more reliable multi‑namespace behavior.
  * More robust EvalHub URL resolution and failure-reporting flow.

* **Bug Fixes / Permissions**
  * Expanded manager RBAC to allow reading/patching workloads and reading Jobs and EvalHub resources.

* **Tests**
  * Added unit tests for workload-failure detection, Job owner extraction, EvalHub job validation, and updated RBAC expectations.


controllers/evalhub/evalhub_tenant_namespaces.go
```diff
@@ -0,0 +1,60 @@
+/*
```
mariusdanciu: AFAIK previously the controller was watching for namespaces with the evalhub label. Why do we need to cache the namespaces ?

nbs-rh: @mariusdanciu if I understand correctly, the existing reconciler watches for jobs/pods in certain states and then filters out the ones that don't belong to evalhub tenant namespace (namespace containing evalhub label). It does not watch directly for namespaces with evalhub label. The cache of EH tenant namespaces helps with this filtering without hitting kube API frequently. 
So the namespace cache existed earlier within the jobs reconciler. This PR refactors it for reuse.

mariusdanciu: I though @ruivieira already had the code for actual watching for namespaces labels as evalhub. I see it in tenant_namespaces.go

mariusdanciu (approved):
lgtm

mariusdanciu (approved):
lgtm

controllers/evalhub/evalhub_tenant_namespaces.go
ruivieira: @mariusdanciu That's right, `tenant_namespaces.go` already watches and lists namespaces with the `evalhub` label.
@nbs-rh IIUC, this is just extracting it so there isn't a duplicate watcher?

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -0,0 +1,278 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta1"
```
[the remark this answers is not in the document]
ruivieira: IMO this should be fine, since it's just registering the type, but not performing any Kueue API operations (i.e. IIUC if the API server doesn't serve `kueue.x-k8s.io/v1beta1`, then no Kueue operations are registered in `kueue_workloads_discovery.go`). I think in this case the operator can work fine if there is no Kueue in the cluster.

ruivieira (approved):

controllers/evalhub/evalhub_tenant_namespaces.go
nbs-rh: @ruivieira right, we have different reconcilers doing different things - there is no duplicate/redundant code.

PR #713, opened 2026-04-20
feat(evalhub): add ServiceMonitor auto-discovery for Prometheus metrics

ruivieira:
## What and why

EvalHub exposes a `/metrics` endpoint but has no `ServiceMonitor`, making it invisible to OpenShift's Prometheus Operator. This is a silent observability gap affecting every EvalHub deployment.

This PR adds three coordinated changes to the operator's EvalHub reconciliation loop:
- **ServiceMonitor** targeting the EvalHub Service on the HTTPS metrics port with owner-reference lifecycle management
- **NetworkPolicy** permitting Prometheus ingress on the metrics port
- **Dynamic namespace-scoped RBAC** (Role + RoleBinding) so the operator only holds monitoring permissions where EvalHub is deployed

Graceful degradation when `monitoring.coreos.com/v1` CRD is absent.

Closes [RHAISTRAT-1507](https://redhat.atlassian.net/browse/RHAISTRAT-1507)

## Type

- [x] feat
- [ ] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [ ] Tests added or updated
- [ ] Tested manually
- [ ] Verified ServiceMonitor target shows `UP` in OCP Prometheus
- [ ] Verified cleanup on EvalHub CR deletion

## Breaking changes

None.

## Release Notes

* **New Features**
  * Added ServiceMonitor support for automatic Prometheus metrics scraping from EvalHub instances
  * Monitoring status tracked via `MonitoringReady` and `ServiceMonitorFailed` conditions
  * Automatic feature detection with graceful degradation on unsupported clusters

* **Chores**
  * Updated test configuration and RBAC permissions for monitoring support

PR #724, opened 2026-05-06
fix(evalhub): prevent reconcile-loop thrashing from ServiceMonitor status updates (#713 follow-up)

ruivieira:
## What and why                                                                                                                                                                   
                                                            
PR #713 introduced ServiceMonitor auto-discovery for EvalHub. Two issues in that PR caused a tight reconcile loop (~3s cycle) that prevented the EvalHub deployment from reaching ready state, timing out all NetworkPolicy and ServiceMonitor ODH tests:
                                                                                                                                                                                 
1. `reconcileServiceMonitor()` unconditionally called `r.Update()` on every reconcile, even when the ServiceMonitor spec hadn't changed. Since the ServiceMonitor spec is derived entirely from immutable CR fields (name, namespace), it never needs to be updated in place, create-only is sufficient.
2. Mid-reconcile `r.Status().Update()` calls on both the success and error paths of ServiceMonitor reconciliation bumped the EvalHub CR's resourceVersion on every loop, triggering immediate re-reconciles that bypassed the 30-second `RequeueWithDelay`.

Refer to [RHOAIENG-61189](https://redhat.atlassian.net/browse/RHOAIENG-61189)

## Changes                                                                                                                                                                        
                                                            
- `servicemonitor.go`: Switch to create-only. Ff the ServiceMonitor already exists, return nil. The ServiceMonitor is garbage-collected via owner reference when the CR is deleted, so no update path is needed.                     
- `evalhub_controller.go`: Remove mid-reconcile `r.Status().Update()` calls from the monitoring block. The `MonitoringDegraded` condition is still set in memory via `instance.SetStatus()` and persisted by the single `updateStatus()` call at the end of the reconcile loop.                                                                         
- `unit_test.go`: Add `TestEvalHubReconciler_reconcileServiceMonitor_NoOpUpdate` . Verifies that calling `reconcileServiceMonitor` twice with the same inputs produces zero `Update` calls.                                                                                                                                                                         
                                                            
## Type                                                                                                                                                                           
                                                            
- [x] fix                                              
                                                            
## Testing

- Tests added or updated                                                                                                                                                       
- Tested manually
                                                                                                                                                                                 
Unit tests: `go test -v ./controllers/evalhub/...` all pass.                                                                                                                   
  
Cluster validation: confirmed via operator logs that the ServiceMonitor is created once (no "Updating ServiceMonitor" lines), the deployment reaches Ready within ~17s, and the reconcile loop settles to the expected 5-minute interval after reaching steady state.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Refactor**
  * Optimized ServiceMonitor reconciliation to defer status updates instead of making mid-reconcile API calls, improving performance and stability.
  * Removed in-place update logic for ServiceMonitor; now created once based on immutable configuration fields.

* **Tests**
  * Added test coverage for ServiceMonitor reconciliation to verify creation and no-op behavior on subsequent calls.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/controllers/evalhub/evalhub_controller.go
+++ b/controllers/evalhub/evalhub_controller.go
@@ -221,14 +221,15 @@ func (r *EvalHubReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	// Reconcile monitoring resources (ServiceMonitor).
 	// Monitoring failures are non-fatal: log, set a degraded condition, and continue.
+	// The condition is set in memory only — updateStatus() at the end of the loop
+	// persists the full status in a single write, avoiding mid-reconcile status
+	// updates that would trigger unnecessary re-reconciles.
 	if r.isServiceMonitorSupported() {
 		if err := r.reconcileServiceMonitor(ctx, instance); err != nil {
 			log.Error(err, "Failed to reconcile ServiceMonitor")
 			instance.SetStatus("MonitoringDegraded", "ServiceMonitorFailed", fmt.Sprintf("Failed to reconcile ServiceMonitor: %v", err), corev1.ConditionTrue)
-			r.Status().Update(ctx, instance)
 		} else {
 			instance.SetStatus("MonitoringDegraded", "MonitoringReady", "", corev1.ConditionFalse)
-			r.Status().Update(ctx, instance)
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


ppadashe-psp (approved):
lgtm

PR #727, opened 2026-05-08
feat(evalhub): add offline flag to CR and config.yaml

scheruku-rh:
**Summary**

- Add optional spec.offline to the EvalHub CRD.
- When spec.offline: true, include offline: true in the generated config.yaml ConfigMap (<evalhub>-config).
- Update sample manifest and add unit tests covering the new field.

## Release Notes

* **New Features**
  * Added offline mode support for EvalHub. Users can now enable offline operation by setting `spec.offline: true` in the EvalHub custom resource configuration.

### DIFF
--- a/controllers/evalhub/configmap.go
+++ b/controllers/evalhub/configmap.go
@@ -3,6 +3,7 @@ package evalhub
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


controllers/evalhub/configmap.go
```diff
@@ -196,6 +197,11 @@ func (r *EvalHubReconciler) generateConfigData(ctx context.Context, instance *ev
 		},
 	}
 
+	// Offline mode: only render when explicitly true.
+	if instance.Spec.Offline != nil && *instance.Spec.Offline {
```
ruivieira: @scheruku-rh could this be extracted as a `IsOffline()` method on the `EvalHubSpec` and replaced here?
e.g.
```go
config.Offline = instance.Spec.IsOffline()
```

PR #732, opened 2026-05-12
feat: classify Kueue GPU admission failures with user-facing messages

nbs-rh:
Related to https://redhat.atlassian.net/browse/RHAIRFE-2171

When a Kueue workload is inadmissible (QuotaReserved=False/Inadmissible), the reconciler now distinguishes GPU quota exhaustion from generic queue errors by inspecting the Job's pod spec and the Kueue condition message.

GPU jobs that can't be admitted get message_code=gpu_unavailable with a human-readable explanation; all other admission failures use queue_error.

This avoids surfacing raw cluster internals through the eval-hub API.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admission failures now differentiate GPU unavailability from generic quota/queue errors; users see specific failure messages and logs include the selected failure code for easier troubleshooting.
* **Tests**
  * Added coverage verifying GPU-request detection and admission-failure classification across GPU and non-GPU scenarios, including quoted resource names and empty cases.


ruivieira:
@coderabbitai review

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -144,6 +154,73 @@ func jobOwnerFromWorkload(wl *kueue.Workload) (name string, uid types.UID, ok bo
...
+			return true
+		}
+	}
+	return false
+}
+
+// kueueConditionMentionsGPU returns true when the Kueue inadmissible condition message references
+// a GPU resource by name (e.g. "nvidia.com/gpu"). This lets us distinguish GPU-specific quota
+// failures from other admission failures without parsing structured data out of free-form text.
+func kueueConditionMentionsGPU(msg string) bool {
```
ruivieira, line 202: @nbs-rh should we also trim `"` and similar? i.e for ` insufficient quota for "nvidia.com/gpu"` and similar, which would give contradictory results

ruivieira (approved):
@nbs-rh LGTM, thanks.

Just a couple of minor comments. One is inline, the other one isthat the file permissions changed from `100644` to `100755`, since the Go files don't need to be executable, is ok to revert this? Thanks!

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
nbs-rh: @ruivieira I did some quick local test and all the below were detected to be containing a GPU product string. The third scenario is the message containing quoted GPU product name. Is this a good test of the scenario you had in mind, or am I missing something?
```
[OK] mentionsGPU("insufficient quota for nvidia.com/gpu in flavor default") = true (want true)
[OK] mentionsGPU("insufficient quota for amd.com/gpu in cluster queue") = true (want true)
[OK] mentionsGPU("insufficient quota for \"nvidia.com/gpu\"") = true (want true)
[OK] mentionsGPU("ClusterQueue foo is stopped") = false (want false)
[OK] mentionsGPU("") = false (want false)
```
I have also added a unit test for the scenario of quoted product name in the message.

nbs-rh: Also, the inadvertent change to file permissions has been reverted. Thanks!

PR #745, opened 2026-06-01
fix(evalhub): fix kube-rbac-proxy auth mappings and switch upstream to HTTP loopback

ruivieira:
## What and why

Two fixes for the EvalHub deployment that prevented the kube-rbac-proxy from starting and routing traffic correctly.

1. Missing methods in `kube-rbac-proxy` auth config (`configmap.go`)

The generated `auth.yaml` had three endpoint mappings without a methods field. The ODH `kube-rbac-proxy` rejects any auth config where a mapping does not declare a non-empty methods list, crashing at startup with:

```
  command failed: invalid authorization config: mappings[N] must specify a non-empty methods list
```

Added explicit methods to the catch-all mapping on `/api/v1/evaluations/jobs` (`[get, delete, put, patch]`) and to the single mappings on `/api/v1/evaluations/collections` and `/api/v1/evaluations/providers` (`[get, post, delete, put, patch]`).

2. EvalHub serving HTTPS while proxy upstream was HTTP (`deployment.go`, `route.go`)

The `kube-rbac-proxy` was configured with `--upstream=http://127.0.0.1:8444/` but EvalHub was starting in TLS mode (it enables TLS whenever `TLS_CERT_FILE`/`TLS_KEY_FILE` are set). This caused a continuous stream of TLS handshake errors on the loopback and the pod never became healthy from the router's perspective.

Removed `TLS_CERT_FILE` and `TLS_KEY_FILE` from EvalHub's container env so it serves plain HTTP on the loopback, matching the proxy's upstream scheme. The `kube-rbac-proxy` continues to handle all external TLS on port 8443 via its own service cert. This follows the same pattern as TAS. Port 8444 is not exposed in the Service, so all traffic (internal and external) must pass through the proxy and its SAR checks.

## Type

- [x] fix

## Testing

- [x] Tested manually
- Authenticated GET /api/v1/evaluations/providers returns 200 with provider list
- Unauthenticated request returns non-2xx (SAR enforced by proxy)
- Route health check (/api/v1/health) responds correctly via the public route

## Summary by CodeRabbit

* **Configuration**
  * Authorization configurations now explicitly define HTTP methods for API endpoint mappings.
  * Updated deployment environment variables: added PORT setting, removed TLS certificate and key variables.

### DIFF
--- a/controllers/evalhub/configmap.go
+++ b/controllers/evalhub/configmap.go
@@ -304,7 +304,8 @@ func generateAuthConfigData() string {
                 apiGroup: mlflow.kubeflow.org
                 resource: experiments
                 verb: get
-        - resources:
+        - methods: [get, delete, put, patch]
+          resources:
             - rewrites:
                 byHttpHeader:
                   name: X-Tenant
@@ -315,7 +316,8 @@ func generateAuthConfigData() string {
                 verb: "{{.FromMethod}}"
     - path: /api/v1/evaluations/collections
       mappings:
-        - resources:
+        - methods: [get, post, delete, put, patch]
+          resources:
             - rewrites:
                 byHttpHeader:
                   name: X-Tenant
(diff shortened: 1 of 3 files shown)


ppadashe-psp:
@nbs-rh  Can you please review the changes ?

julpayne (approved):
LGTM

nbs-rh (approved):
LGTM. Thanks!

PR #750, opened 2026-06-06
refactor(evalhub): reuse evalHubImage for MCP deployment instead of separate key

ruivieira:
## What and why

The MCP deployment was using a separate `evalHubMCPImage` ConfigMap key to resolve its container image, falling back to `quay.io/evalhub/evalhub:latest`.
Since the MCP server binary (`evalhub-mcp`) is shipped in the same image as the EvalHub service, this required operators to set two identical image entries in `params.env` or accept the public fallback image instead of a locally-built one.

Simplify by reusing the existing `evalHubImage` ConfigMap key for the MCP deployment. The CR-level `spec.mcp.image` field still provides a per-instance override when a genuinely different image is needed.

Closes [RHOAIENG-66841](https://redhat.atlassian.net/browse/RHOAIENG-66841)

## Type

- [ ] feat
- [ ] fix
- [x] refactor / chore
- [ ] docs
- [ ] test / ci

## Testing

- [x] Tested manually

After operator redeploy with only `evalHubImage` set in `params.env`, the MCP pod uses the correct internal registry image:

```  
oc get pod -n opendatahub -l  -o jsonpath='{..image}'
```

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal code organization for MCP component configuration and image handling.

### DIFF
--- a/controllers/evalhub/constants.go
+++ b/controllers/evalhub/constants.go
@@ -86,10 +86,8 @@ const (
 	sidecarBaseURL = "http://localhost:8080"
 
 	// MCP server configuration
-	defaultMCPImage      = "quay.io/evalhub/evalhub:latest"
-	configMapMCPImageKey = "evalHubMCPImage"
-	mcpBinaryPath        = "/app/evalhub-mcp"
-	mcpContainerName     = "evalhub-mcp"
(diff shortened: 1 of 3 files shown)


ppadashe-psp (approved):
lgtm

PR #759, opened 2026-06-11
feat(evalhub): grant eval-hub SA permissions to manage model credential secrets

scheruku-rh:
https://redhat.atlassian.net/browse/RHOAIENG-53360

During eval job submission, the eval-hub service creates an ephemeral internalModelRef secret in the tenant namespace containing placeholder ref tokens (e.g. api-key:ref) instead of real credentials. This secret is mounted into the adapter container; the sidecar intercepts model calls, swaps the ref token for the real credential, and forwards to the model endpoint. The secret is garbage-collected via Job owner reference when the job completes.
This PR grants the eval-hub service SA the RBAC permissions needed to manage this secret lifecycle.

## Summary by CodeRabbit

* **Chores**
  * Added RBAC permissions for tenant-scoped model credential secrets, including ephemeral ref-token secrets created during job submission.
  * Introduced a dedicated model-secret ClusterRole and a tenant RoleBinding as part of EvalHub job service account setup.
  * Updated controller/manager RBAC and policy enforcement to allow only the required `clusterrole bind` and secret-write exemptions.
  * Added RBAC policy tests covering the new exemption behavior.


julpayne (approved):
LGTM

controllers/evalhub/evalhub_controller.go
```diff
@@ -65,6 +65,7 @@ type EvalHubReconciler struct {
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch;update
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=rolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;create;delete
```
ruivieira: @scheruku-rh  just one question:

The `get/create/delete` on secrets is added to the manager's `ClusterRole`, which is bound cluster-wide. The actual secret lifecycle is handled by the eval-hub service SA (via the new RoleBinding). So the manager never touches secrets directly. My guess is the marker is there to satisfy Kubernetes' privilege escalation check (you can't create a `RoleBinding` granting permissions you don't already hold). Is that right?

If so, IMO a better alternative would be to grant `bind` on the specific `ClusterRole` instead of the raw secret verbs cluster-wide:

```yaml
  - apiGroups: ["rbac.authorization.k8s.io"]
    resources: ["clusterroles"]
    resourceNames: ["trustyai-service-operator-evalhub-model-secret"]
    verbs: ["bind"]
```

With the current approach the manager pod can directly read, create, and delete any secret in any namespace. With `bind`, the manager can only delegate that access to the service SA (it can't touch secrets itself at all).

@coderabbitai do you agree?

ruivieira (changes requested):

ruivieira (approved):

PR #760, opened 2026-06-11
feat(evalhub): inject discovery ConfigMap into tenant namespaces for service URL resolution

ruivieira:
## What and why

Adds a well-known `evalhub-discovery` ConfigMap that is injected into every tenant namespace during reconciliation. The ConfigMap holds one key per EvalHub instance (`{instanceName}.url`) pointing to its in-cluster service URL, so clients in tenant namespaces can resolve the EvalHub endpoint without cross-namespace API lookups.

Multiple EvalHub instances co-exist safely. Each owns exactly one key in the shared ConfigMap. When a namespace loses the tenant label, the instance's key is removed; if that empties the ConfigMap, it is deleted.

Closes [RHOAIENG-68253](https://redhat.atlassian.net/browse/RHOAIENG-68253)

## Type

- [x] feat
- [ ] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [x] Tests added or updated
- [x] Tested manually

Unit tests cover:
- ConfigMap creation in a new tenant namespace
- URL drift detection and update
- Multi-instance key coexistence (own key added, other keys preserved)
- Idempotency (no spurious updates when URL is unchanged)
- Cleanup: CM deleted when it becomes empty, only own key removed when others remain, active-tenant CMs left untouched

## Breaking changes

None.

## Release Notes

* **New Features**
  * EvalHub instances now automatically register and discover each other's service URLs within tenant namespaces
  * Service location information is automatically maintained and cleaned up to remove stale entries, ensuring accurate routing

* **Tests**
  * Added comprehensive unit tests for service discovery and cleanup functionality

### DIFF
--- a/controllers/evalhub/constants.go
+++ b/controllers/evalhub/constants.go
@@ -98,6 +98,12 @@ const (
 	mcpServiceCAVolumeName = "mcp-service-ca"
 	mcpServiceCAMountPath  = "/etc/evalhub-mcp/ca"
 
+	// Discovery ConfigMap injected into every tenant namespace for EvalHub service URL resolution.
+	// A single well-known CM is shared across all EvalHub instances; each instance owns one
+	// key of the form "{instanceName}.url" so multiple EvalHub instances can coexist.
+	discoveryConfigMapName  = "evalhub-discovery"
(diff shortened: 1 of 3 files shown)


julpayne (approved):
LGTM

PR #770, opened 2026-06-16
fix: replace verbose status messages with concise ones

nbs-rh:
This fixes an error reported by the UI team. evaluation_job_failure_reconciler earlier sent the entire error message that it received from kube. With this fix, if the message is longer than a threshold, we look for specific substrings in the message and replace it with a shorter/concise message for the API caller.

### Tests

# Error Message Sanitization - Test Summary Report

## Test Execution Results

**Date**: June 16, 2026
**Status**: ✅ **ALL TESTS PASSING**
**Test Suite**: `controllers/evalhub/sanitize_standalone_test.go`
**Total Tests**: 16 (7 main tests + 9 sub-tests)
**Pass Rate**: 100%

## Summary by CodeRabbit

* **Bug Fixes**
  * Failure notifications now show cleaner, shorter error messages for common image-pull and connectivity issues.
  * Very long errors are automatically shortened to keep messages readable.
  * Error matching is more consistent across different text casing and formats.

* **Chores**
  * Updated internal configuration and permissions settings.


controllers/evalhub/evaluation_job_failure_reconciler.go
```diff
@@ -549,6 +552,88 @@ func isFatalContainerWaitingReason(reason string) bool {
...
+		}
+	}
+
+	// If no pattern matched and message is short enough, return as-is
+	if len(message) <= maxErrorMessageLength {
+		return message
+	}
+
+	// Fallback: truncate with ellipsis indicating there's more detail in pod logs
+	return message[:maxErrorMessageLength-3] + "..."
```
ruivieira: IIRC  Go can panic if the slice boundary falls inside a multi-byte rune. K8s messages are ASCII, but if a URL, image name, or error has unicode, there's a risk the operator will crash.

Can we use something like
```
  runes := []rune(message)
  if len(runes) > maxErrorMessageLength {
      return string(runes[:maxErrorMessageLength-3]) + "..."
  }
  return message
```
Alternatively we could convert before with `utf8.ValidString` and `strings.ToValidUTF8`
Thanks!

@coderabbitai do you agree?

ruivieira (approved):

PR #774, opened 2026-06-18
feat: integrate with cluster TLS security profile

ugiordan:
## Summary

- Read the cluster TLS profile from `apiservers.config.openshift.io/cluster` at startup
- TLS resolution code extracted into `pkg/tls/` for reusability and testability
- `pkg/tls.Resolve()` uses unstructured access (zero new dependencies, compatible with controller-runtime v0.17.0)
- Apply MinVersion, CipherSuites, and NextProtos to metrics server TLS config
- Fail closed on unexpected errors, use Intermediate defaults on non-OpenShift clusters
- Transient API errors (ServiceUnavailable, Timeout, TooManyRequests) fall back to Intermediate defaults instead of crashing
- 10s context timeout on APIServer fetch to avoid blocking startup indefinitely
- Add `config.openshift.io/apiservers` to OPA policy allowlist
- Add RBAC ClusterRole and ClusterRoleBinding for reading APIServer CR
- ClusterRoleBinding and ClusterRole registered in `config/rbac-base/kustomization.yaml`
- CRB added to OPA `policy/rbac.rego` allowlist (both prefixed and un-prefixed)

## Files changed

- `cmd/main.go`: Replaced inline TLS code with call to `pkgtls.Resolve()`
- `pkg/tls/tls.go`: Reusable TLS profile resolution (unstructured APIServer read, cipher mapping, Intermediate defaults, transient error handling, 10s timeout)
- `pkg/tls/tls_test.go`: Table-driven tests for all profile types
- `config/rbac-base/tls_profile_role.yaml`: RBAC ClusterRole for reading APIServer
- `config/rbac-base/tls_profile_role_binding.yaml`: RBAC ClusterRoleBinding for operator SA
- `config/rbac-base/kustomization.yaml`: Added TLS RBAC resources
- `policy/clusterrole.rego`: OPA allowlist update for `config.openshift.io/apiservers`
- `policy/rbac.rego`: OPA allowlist update for TLS CRB (prefixed + un-prefixed)
- `Dockerfile`: Added `COPY pkg/ pkg/` for container builds

## Test plan

- [x] Unit tests pass (`go test ./pkg/tls/... -v`)
- [x] Conftest RBAC policy passes
- [ ] Operator starts on OpenShift with Intermediate profile
- [ ] Operator starts on non-OpenShift with hardened defaults

Ref: [RHOAIENG-61068](https://redhat.atlassian.net/browse/RHOAIENG-61068)


ugiordan:
@ruivieira, @RobGeada, could you please review this PR?

cmd/main.go
ruivieira: @ugiordan don't we need to propagate tlsOpts to the EvalHub webhook too?

config/rbac-base/tls_profile_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
ruivieira, line 1: @ugiordan Should we add to a rolebinding to the operator's service account?
Then we need to add both (this and the rolebinding) to config/rbac-base/kustomization.yaml?

pkg/tls/tls.go
```diff
@@ -0,0 +1,171 @@
...
+
+	k8sClient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return result, fmt.Errorf("creating bootstrap client for TLS profile: %w", err)
+	}
+
+	apiServer := &configv1.APIServer{}
+	if err := k8sClient.Get(ctx, client.ObjectKey{Name: "cluster"}, apiServer); err != nil {
+		switch {
+		case meta.IsNoMatchError(err):
```
ruivieira, line 95: Should we add `apierrors.IsServiceUnavailable` and `apierrors.IsTimeout`/`apierrors.IsTooManyRequests` to the graceful cases?

ruivieira (changes requested):
@ugiordan thanks for the PR.

On top of the inlined comments, I want to ask if it's possible to add some retry mechanism.

`Resolve` is called once at process start before the manager is initialised, so there is no controller-runtime retry loop or leader-election backoff for transient API server errors. A single network timeout at pod start propagates as a non-nil error and causes `os.Exit(1)` in `main.go`. So what's a minor temporary network issue can turn into a `CrashLoopBackOff`.

Since the fallback to Intermediate defaults is already implemented for the permanent cases (`NoMatch`, `NotFound`, `Forbidden`), transient errors should also reach that fallback rather than crash the process.

If a short exponential backoff (e.g. 5 steps, 0.5s base, factor 2and ~15s ceiling) on `IsServiceUnavailable`, `IsTimeout`, and `IsTooManyRequests` before falling back covers this window without blocking normal startup.

wdyt?

ugiordan:
> @ugiordan thanks for the PR.
> 
> On top of the inlined comments, I want to ask if it's possible to add some retry mechanism.
> 
> `Resolve` is called once at process start before the manager is initialised, so there is no controller-runtime retry loop or leader-election backoff for transient API server errors. A single network timeout at pod start propagates as a non-nil error and causes `os.Exit(1)` in `main.go`. So what's a minor temporary network issue can turn into a `CrashLoopBackOff`.
> 
> Since the fallback to Intermediate defaults is already implemented for the permanent cases (`NoMatch`, `NotFound`, `Forbidden`), transient errors should also reach that fallback rather than crash the process.
> 
> If a short exponential backoff (e.g. 5 steps, 0.5s base, factor 2and ~15s ceiling) on `IsServiceUnavailable`, `IsTimeout`, and `IsTooManyRequests` before falling back covers this window without blocking normal startup.
> 
> wdyt?

@ruivieira makes sense. Added `IsServiceUnavailable`, `IsTimeout`, and `IsTooManyRequests` to the graceful fallback cases in the latest push. Transient API errors at startup now fall back to Intermediate defaults instead of crashing.

I went with graceful fallback over exponential backoff to keep startup simple. The `SecurityProfileWatcher` will pick up the correct profile once the manager starts and the API server is reachable again.

pkg/tls/tls.go
ugiordan: Done.

config/rbac-base/tls_profile_role.yaml
ugiordan: Good catch. Added the `ClusterRoleBinding` and registered both the `ClusterRole` and `ClusterRoleBinding` in `config/rbac-base/kustomization.yaml`.

cmd/main.go
ugiordan: fixed

ugiordan:
Hey @ruivieira, unfortunately, this PR got into a broken state (disconnected git history from a shallow clone mishap on my side), so I had to open a replacement: https://github.com/trustyai-explainability/trustyai-service-operator/pull/812

PR #775, opened 2026-06-18
fix(evalhub): create discovery ConfigMap in instance namespace for non-admin users

ruivieira:
## What and why

Non-admin users in the namespace where the EvalHub CR lives cannot resolve the service URL. The controller was skipping the instance namespace entirely when creating the evalhub-discovery ConfigMap, assuming users there could read the CR directly. But non-admin users only have virtual resource RBAC (evaluations, collections, providers). They can't list evalhubs.trustyai.opendatahub.io, so the BFF gets a 403 and returns a 500.

## Type

- [ ] feat
- [x] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [x] Tests added or updated
- [ ] Tested manually

## Breaking changes

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed service URL discovery for non-admin users by ensuring discovery configuration is provisioned in instance namespaces.

* **Tests**
  * Updated tests to verify discovery configuration provisioning in instance namespaces.

### DIFF
--- a/controllers/evalhub/tenant_namespaces.go
+++ b/controllers/evalhub/tenant_namespaces.go
@@ -39,8 +39,14 @@ func (r *EvalHubReconciler) reconcileTenantNamespaces(ctx context.Context, insta
 	}
 
 	for ns := range activeTenants {
-		// Skip the instance namespace (already handled by main reconcile)
+		// The instance namespace already has SA and RBAC from the main reconcile,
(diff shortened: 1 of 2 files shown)


ppadashe-psp (approved):
lgtm

kpunwatk (approved):

ruivieira:
This PR is being closed after architectural clarification with the stakeholders.

PR #776, opened 2026-06-18
Grant EvalHub MCP API access via instance-namespace Role rules

julpayne:
## Summary
- Extend the per-instance `{name}-service-access-role` with virtual SAR resources for `evaluations`, `providers`, and `collections` (full get/list/create/update/patch/delete verbs) so the EvalHub/MCP service account can access EvalHub REST endpoints from the instance namespace.
- Stop creating legacy `{name}-providers-access-rb` and `{name}-collections-access-rb` ClusterRole bindings for the service SA.
- Delete those legacy RoleBindings on reconcile when upgrading existing instances.

## Test plan
- [x] `go test ./controllers/evalhub/... -run 'TestEvalHubReconciler_createAPIAccessRole|EvalHub API RBAC'`
- [ ] Verify MCP `resources/read` for `evalhub://collections` succeeds when `EVALHUB_TENANT` matches the EvalHub instance namespace and `authSecret` contains the `{name}-service` token

## Release Notes

* **Chores**
  * Streamlined EvalHub access control by removing outdated legacy bindings and configurations.
  * Enhanced resource-specific access permissions for evaluations, providers, and collections through refined API rules.

### DIFF
--- a/controllers/evalhub/proxy_rbac_test.go
+++ b/controllers/evalhub/proxy_rbac_test.go
@@ -163,10 +163,16 @@ var _ = Describe("EvalHub API RBAC", func() {
 			}, role)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("Checking Role has resourceNames set to instance name")
-			Expect(role.Rules).To(HaveLen(2))
+			By("Checking Role has resourceNames set to instance name and virtual API resources")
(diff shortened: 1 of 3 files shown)


julpayne:
Not valid.

PR #797, opened 2026-06-30
ci: add disconnected readiness scanner GHA workflow

ruivieira:
## What and why

Adds a GitHub Actions workflow that runs the disconnected readiness scanner on every push and PR. Catches disconnected-breaking patterns at PR time rather than later.

## Type

- [ ] feat
- [ ] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci
- [x] test / ci

## Testing

- [ ] Tests added or updated
- [x] Tested manually

## Summary by CodeRabbit

* **New Features**
  * Added a new automated readiness check that runs on pushes and pull requests.
  * Expanded continuous integration coverage to validate disconnected-readiness status for changes.

PR #798, opened 2026-06-30
fix(kustomize): replace install script with direct binary download

sheltoncyril:
## Summary
- Replace `install_kustomize.sh` script with direct binary download from GitHub releases
- The script hits `api.github.com` unauthenticated to resolve release URLs, which gets rate-limited (60 req/hr per shared runner IP) — root cause of frequent CI failures
- Direct download URL is deterministic and bypasses the API entirely

**Depends on:** #797

## Test plan
- [x] Verified `rm -f bin/kustomize && make kustomize && bin/kustomize version` succeeds locally
- [ ] CI workflows that use `make kustomize` pass without rate-limit errors

## Summary by CodeRabbit

* **Chores**
  * Updated the local tooling setup to fetch the required `kustomize` binary directly from the official release package.
  * Improved platform detection so the correct download is used for common CPU architectures.
  * Existing installs still verify the expected version and are refreshed when needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,6 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 KUSTOMIZE_VERSION ?= v5.7.0
 CONTROLLER_TOOLS_VERSION ?= v0.20.0
 
-KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
 $(KUSTOMIZE): $(LOCALBIN)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -244,15 +244,19 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
...
 $(KUSTOMIZE): $(LOCALBIN)
 	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
 		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
 		rm -rf $(LOCALBIN)/kustomize; \
 	fi
-	test -s $(LOCALBIN)/kustomize || { curl -Ss $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); }
+	@test -s $(LOCALBIN)/kustomize || { \
+		OPSYS=$$(uname -s | tr '[:upper:]' '[:lower:]') && \
+		ARCH=$$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/') && \
+		curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_$${OPSYS}_$${ARCH}.tar.gz" | \
```
ruivieira: @sheltoncyril could we add `--fail` here?

ruivieira (approved):

Makefile
sheltoncyril: Done, its `curl -sfL ...` now, thank you!

PR #813, opened 2026-07-07
fix(evalhub): grant service SA pod log access in tenant namespaces

julpayne:
## Summary

- Add a scoped namespace Role (`pods` get/list, `pods/log` get) and RoleBinding so the EvalHub service SA can read job pod logs in tenant namespaces
- Reconcile pod-log RBAC alongside existing tenant-namespace resources (job SA, jobs-writer, job-config, etc.) so existing clusters are backfilled on operator upgrade
- Also provision pod-log RBAC in the EvalHub instance namespace for single-tenant deployments

Fixes the HTTP 500 on `GET /api/v1/evaluations/jobs/{job_id}/logs` caused by:
`pods is forbidden: User "system:serviceaccount:evalhub:evalhub-service" cannot list resource "pods"`.

## Test plan

- [x] `go test -v ./controllers/evalhub/... -run TestEvalHubReconciler_reconcileTenantNamespaces`
- [ ] After deploying operator upgrade, verify on cluster:
  - `kubectl auth can-i list pods -n <tenant> --as=system:serviceaccount:evalhub:evalhub-service` → yes
  - `kubectl auth can-i get pods/log -n <tenant> --as=system:serviceaccount:evalhub:evalhub-service` → yes
  - `GET /api/v1/evaluations/jobs/{job_id}/logs` returns 200 with log text
  - `uv run python examples/watch_job_logs.py <job_id> --profile <profile>` streams logs

## Summary by CodeRabbit

* **New Features**
  * EvalHub now configures RBAC so the instance service account can read job pods and retrieve their logs in the target namespace.
  * Active tenant namespaces now receive the same pod/log access so the instance API can read logs there too.
* **Bug Fixes**
  * Improved RBAC provisioning to create the required Roles and RoleBindings and keep the permissions in sync.


controllers/evalhub/service_accounts.go
```diff
@@ -195,6 +202,21 @@ func (r *EvalHubReconciler) createServiceAccount(ctx context.Context, instance *
 		return err
 	}
 
+	err = r.createServicePodLogsRole(ctx, instance, instance.Namespace)
+	if err != nil {
+		return err
+	}
+
+	podLogsRBName := normalizeDNS1123LabelValue(instance.Name + "-service-pod-logs-rb")
```
ruivieira: @julpayne could this include the namespace too? For instance

```go
podLogsRBName := normalizeDNS1123LabelValue(instance.Name + "-" + instance.Namespace + "-service-pod-logs-rb")
```

If two multi-tenant EH instances target the same namespace, these would colide.

This is _not_ supported, but at least it would avoid breaking RBAC if someone decides to do it

controllers/evalhub/unit_test.go
```diff
@@ -2202,6 +2202,62 @@ func TestEvalHubReconciler_reconcileTenantNamespaces(t *testing.T) {
 		assert.Equal(t, instanceNamespace, rb.Subjects[0].Namespace)
 	})
 
+	t.Run("should create service pod-logs Role and RoleBinding in tenant namespace", func(t *testing.T) {
```
ruivieira, line 2205: @julpayne should we add a test that calls `createServiceAccount`, then asserts the `Role` and `RoleBinding` exist in `instance.Namespace` with the correct names and rules?

ruivieira (changes requested):
Thanks for the PR @julpayne!

LGTM, but Just left a couple of questions.

controllers/evalhub/service_accounts.go
julpayne: Done

controllers/evalhub/unit_test.go
julpayne: Done

ruivieira (approved):

PR #821, opened 2026-07-11
fix(rbac,lmes): correct TLS profile role binding subject and harden device detection test

ruivieira:
## What and why

Two unrelated fixes bundled together:

1. **RBAC**: The `tls-profile-reader` RoleBinding referenced `trustyai-service-operator` as the ServiceAccount subject, but the actual manager ServiceAccount is named `controller-manager`. This mismatch prevented the operator from reading TLS profile resources at runtime.

2. **LMES driver test**: `Test_DetectDeviceError` assumed `torch` was never installed in the test environment, causing it to fail on machines where PyTorch is present (the device detection would succeed, making the error-path assertion incorrect). The fix skips the test when `torch` is available, and relaxes the message assertion to use `Contains` rather than `Equal` so minor variations in the error string don't break the test.

## Type

- [x] fix
- [ ] feat
- [ ] docs
- [ ] refactor / chore
- [x] test / ci

## Testing

- [x] Tests added or updated
- [ ] Tested manually

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated TLS profile access permissions to use the correct controller service account, improving reliable certificate profile handling.

- **Tests**
  - Improved device-detection error testing across environments by skipping the test when hardware support is available and using more resilient error-message validation.

### DIFF
--- a/config/rbac-base/tls_profile_role_binding.yaml
+++ b/config/rbac-base/tls_profile_role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: tls-profile-reader
 subjects:
   - kind: ServiceAccount
-    name: trustyai-service-operator
+    name: controller-manager
     namespace: system
--- a/controllers/lmes/driver/driver_test.go
+++ b/controllers/lmes/driver/driver_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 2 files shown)


ruivieira:
cc @ugiordan follow up fix to https://github.com/trustyai-explainability/trustyai-service-operator/pull/812. Could you PTAL, thanks!

ugiordan (approved):

ABeltramo (approved):

PR #825, opened 2026-07-16
fix(evalhub): grant evalhubs CR read access to single-tenancy roles

ruivieira:
## What and why

The BFF cannot discover the EvalHub service URL in single-tenancy mode using the user's token. In multi-tenancy, the operator creates an `evalhub-discovery` ConfigMap in each tenant namespace, but in single-tenancy no such ConfigMap exists: the alternative is reading `Status.URL` from the EvalHub CR directly. Neither `evalhub-user` nor `evalhub-tenant-admin` Roles grant `get` or `list` on the `evalhubs` resource, so the CR lookup fails with a 403.

This adds a read-only (`get`, `list`) rule for `evalhubs` to both single-tenancy convenience Roles. The permission is namespace-scoped (Role, not ClusterRole) and only created when `spec.tenancy: single`.

## Type

- [ ] feat
- [x] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [x] Tests added or updated
- [ ] Tested manually

## Breaking changes

None.

## Summary by CodeRabbit

* **New Features**
  * Tenant administrators and users can now discover EvalHub resources by listing or viewing them within their namespace.

* **Bug Fixes**
  * Updated access permissions so required EvalHub resources are visible in single-tenant environments.

### DIFF
--- a/controllers/evalhub/tenant_roles.go
+++ b/controllers/evalhub/tenant_roles.go
@@ -134,6 +134,11 @@ func tenantAdminRules(instance *evalhubv1.EvalHub) []rbacv1.PolicyRule {
 	return []rbacv1.PolicyRule{
 		{
 			APIGroups: []string{"trustyai.opendatahub.io"},
+			Resources: []string{"evalhubs"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"trustyai.opendatahub.io"},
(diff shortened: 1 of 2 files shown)


saichandrapandraju (approved):

sheltoncyril (approved):
/lgtm

PR #832, opened 2026-07-23
fix(evalhub): update health probe endpoint to /api/v1/health

ruivieira:
## What and why

Updates the EvalHub probe endpoint from `/healthz` to `/api/v1/health` in line with eval-hub/eval-hub#797, which removes the `/healthz` route and makes `GET /api/v1/health` the single unauthenticated liveness/readiness probe endpoint.

The `--ignore-paths` flag passed to kube-rbac-proxy is updated via the same constant, so kubelet probes continue to bypass SAR enforcement.

Also removes the unused `evalHubHealthPath` constant (was defined but never referenced).

## Type

- [ ] feat
- [x] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [ ] Tests added or updated
- [x] Tested manually

## Breaking changes

Requires eval-hub >= the version from eval-hub/eval-hub#797 (which removes `/healthz`). Deploying this operator change against an older eval-hub image will cause probe failures.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the health check endpoint used by system probes.
  - Health checks can now reach the service without requiring identity headers, improving probe reliability.

### DIFF
--- a/controllers/evalhub/constants.go
+++ b/controllers/evalhub/constants.go
@@ -14,11 +14,10 @@ const (
 	// evalHubAppPort is the eval-hub container listen port on loopback (API_HOST=127.0.0.1, PORT in deployment env).
 	// kube-rbac-proxy upstream is http://127.0.0.1:<evalHubAppPort>/ on the pod network (TLS is terminated on servicePort).
 	evalHubAppPort = 8444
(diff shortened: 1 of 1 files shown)


julpayne (approved):
LGTM

PR #841, opened 2026-07-31
fix(config): add missing kustomize vars for deepeval and ragas provider images

ruivieira:
## Summary

The `evalhub-provider-deepeval-image` and `evalhub-provider-ragas-image` variables were referenced as `$(...)` in their respective provider ConfigMaps but were never declared in the base `kustomization.yaml` vars block. Kustomize would leave the literal placeholder strings in the deployed ConfigMaps, causing any DeepEval or RAGAS evaluation job to fail at pod scheduling time with an invalid image reference.

Additionally, `evalhub-provider-ragas-image` was absent from `config/base/params.env`, meaning the base overlay had no fallback value at all.

## Changes

- `config/base/kustomization.yaml`: add `vars` entries for `evalhub-provider-deepeval-image` and `evalhub-provider-ragas-image`
- `config/base/params.env`: add `evalhub-provider-ragas-image=quay.io/evalhub/community-ragas:latest` as the base default (rhoai and odh overlays already pin this to `v0.5.0` via `behavior: merge`)


m-misiura (approved):

ppadashe-psp (approved):
Approved

PR #843, opened 2026-08-01
chore(rbac): add events:create and jobs:patch/get permissions for EvalHub (#842

ruivieira:
* chore(rbac): add events:create and jobs:patch/get permissions for EvalHub ServiceAccount

Add two new ClusterRoles for the EvalHub ServiceAccount:
- evalhub-events: grants events create, so the EvalHub server can emit Kubernetes Events against backing Job resources on evaluation lifecycle transitions (EvaluationStarted, EvaluationCompleted, EvaluationFailed, EvaluationThresholdViolated)
- evalhub-jobs-writer extended with get and patch verbs, so the server can read-modify-patch the trustyai.opendatahub.io/evaluation-phase label on running Jobs

Both new ClusterRoleBindings are added to the OPA policy allowlist. No functional code changes; this is the RBAC prerequisite for RHAI-277.

* fix(rbac): add patch verb to evalhub-events ClusterRole for EventRecorder deduplication

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EvalHub event reporting by granting permission to create and update event records.
  * Enhanced job management reliability with additional permissions to view and update jobs.
  * Updated access controls to recognize EvalHub event permissions across supported configurations.

PR #846, opened 2026-08-01
feat(evalhub): emit Kubernetes events for operator-detected evaluation failures (RHAI-278)

ruivieira:
## What and why

EvalHub evaluation jobs can fail in ways the EvalHub server cannot self-report OOM kills, image pull errors, and Kueue admission failures all happen before or outside the server's callback path. This PR wires Kubernetes event emission into both failure reconcilers so these infrastructure failures are observable via `kubectl describe job` / `kubectl get events`, completing the failure signal chain for [RHAI-278](https://redhat.atlassian.net/browse/RHAI-278).

> Note: This PR is dependent on https://github.com/trustyai-explainability/trustyai-service-operator/pull/843, which must be merged first.

Closes RHOAIENG-80112, RHOAIENG-80113, RHOAIENG-80114.

## Type

- [x] feat
- [ ] fix
- [ ] docs
- [ ] refactor / chore
- [ ] test / ci

## Testing

- [x] Tests added or updated
- [x] Tested manually

Lifecycle unit tests use `fake.Client`, `record.FakeRecorder`, and a local `httptest.Server` to exercise the full `Reconcile()` path. Cases covered:

- **Dedup — server already handled**: `evaluation-phase=Failed` label already set by EvalHub server → operator returns early, no duplicate event
- **OOM kill**: adapter container terminated with `OOMKilled` → `EvaluationFailed` warning event emitted + `evaluation-phase=Failed` label + `evaluation-status` annotation stamped on Job
- **Image pull error**: init container stuck on `ErrImagePull` → same lifecycle
- **Kueue eviction**: Workload with `QuotaReserved=False / Inadmissible` → `EvaluationFailed` warning event + Job label patch + Workload annotated as reported

## Changes

- **RBAC** (`RHOAIENG-80112`): add `events:create/patch` and `jobs:get/patch` permissions; wire `EventRecorder` into both failure reconciler constructors via `mgr.GetEventRecorderFor`
- **Event emission** (`RHOAIENG-80113`): emit `Warning / EvaluationFailed` after a successful POST to EvalHub; patch `evaluation-phase=Failed` and `evaluation-status` on the Job; Kueue path also annotates the Workload as reported
- **Tests** (`RHOAIENG-80114`): lifecycle unit tests for both reconcilers covering the four failure scenarios above

## Summary by CodeRabbit

* **New Features**
  * Added clearer Kubernetes warning events for evaluation failures.
  * Evaluation failures now update associated Jobs with failed status details.
  * Added permissions required to record events and update Jobs.

* **Bug Fixes**
  * Prevented duplicate handling of failures already reported to EvalHub.
  * Improved handling of admission failures, out-of-memory errors, and image-pull failures.
  * Failed Jobs are cleaned up after successful synchronization.

* **Tests**
  * Added coverage for failure reporting, event emission, status updates, deduplication, and cleanup.


ruivieira:
@coderabbitai review

sheltoncyril (approved):
/lgtm

sheltoncyril (changes requested):
Claude had a few comments

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -281,6 +285,12 @@ func (r *EvalHubEvaluationFailedKueueWorkloadsReconciler) Reconcile(ctx context.
 	if failureAlreadyReported(&job) {
 		return ctrl.Result{}, nil
 	}
+	if serverAlreadyHandledFailure(&job) {
+		log.V(1).Info("skip: EvalHub server already set failure label on Job",
+			append(evaluationFailedKueueWorkloadsLogFields(), "action", "skip_server_handled",
```
sheltoncyril, line 288: **Bug (confirmed): Deadlock on partial commit.**

If `patchJobFailureLabels` succeeds (Job gets `evaluation-phase=Failed`) but `annotateWorkloadReported` fails (API conflict), the retry path hits `serverAlreadyHandledFailure` → early return. The Workload annotation is never written. On every informer resync the Workload is re-enqueued, hits `serverAlreadyHandledFailure`, exits — perpetual no-op reconcile cycles.

Root cause: `serverAlreadyHandledFailure` cannot distinguish the operator's own label (set by `patchJobFailureLabels` a few lines below) from one set by the EvalHub server.

Fix options:
1. Move `serverAlreadyHandledFailure` check **after** `patchJobFailureLabels` (so the operator's own partial-commit label is never mistaken for a server-set label).
2. Use a distinct label value or annotation to mark operator-set failures vs server-set failures.
3. Make `patchJobFailureLabels` + `annotateWorkloadReported` atomic (single patch), so the partial-commit window disappears.

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -325,6 +335,15 @@ func (r *EvalHubEvaluationFailedKueueWorkloadsReconciler) Reconcile(ctx context.
 		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
sheltoncyril, line 346: **Bug (confirmed): Duplicate POST + Event on retry.**

`Eventf` and `patchJobFailureLabels` run before `annotateWorkloadReported`. If `annotateWorkloadReported` fails (and optionally `patchJobFailureLabels` also fails), on retry neither dedup guard is set → reconciler re-POSTs to EvalHub (duplicate HTTP call) and emits a second `EvaluationFailed` Event.

K8s event aggregation mitigates the Event side, but the duplicate POST to EvalHub is not idempotent in the general case.

Consider: move `Eventf` **after** all patches succeed, or make the EvalHub POST idempotent by Job ID.

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -325,6 +335,15 @@ func (r *EvalHubEvaluationFailedKueueWorkloadsReconciler) Reconcile(ctx context.
...
 	}
 
+	r.EventRecorder.Eventf(&job, corev1.EventTypeWarning, eventReasonEvaluationFailed, "%s", failureMsg)
+
+	if err := r.patchJobFailureLabels(ctx, &job, failureMsg); err != nil {
```
sheltoncyril, line 360: **Issue (plausible): `patchJobFailureLabels` failure permanently loses metadata.**

This error is logged but non-fatal — execution continues to `annotateWorkloadReported`. If the annotation succeeds, future reconciles are blocked by `workloadFailedEventAlreadyReported=true`. Job permanently lacks `evaluation-phase=Failed` label and `evaluation-status` annotation.

If `patchJobFailureLabels` is important enough to have, consider making its failure trigger a requeue so the label is retried.

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
```diff
@@ -351,3 +370,18 @@ func (r *EvalHubEvaluationFailedKueueWorkloadsReconciler) annotateWorkloadReport
...
+// patchJobFailureLabels stamps the owning Job with evaluation-phase=Failed and sets the evaluation-status
+// annotation so the failure is observable on the Job resource even before it is cleaned up.
+func (r *EvalHubEvaluationFailedKueueWorkloadsReconciler) patchJobFailureLabels(ctx context.Context, job *batchv1.Job, statusMsg string) error {
+	base := job.DeepCopy()
+	if job.Labels == nil {
+		job.Labels = map[string]string{}
+	}
+	if job.Annotations == nil {
+		job.Annotations = map[string]string{}
+	}
```
sheltoncyril, line 396: **Nit: Unreachable nil-check.**

`isEvalHubEvaluationJob` (called earlier) indexes `job.Labels[evalHubAppLabel]`, which already proves `Labels != nil`. This nil-check can never be true. Same pattern exists in the promote block in `evaluation_job_failure_reconciler.go`.

controllers/evalhub/evaluation_job_failure_reconciler.go
```diff
@@ -409,12 +430,19 @@ func (r *EvalHubEvaluationJobFailureReconciler) Reconcile(ctx context.Context, r
 		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	r.EventRecorder.Eventf(&job, corev1.EventTypeWarning, eventReasonEvaluationFailed, "%s", msg)
```
sheltoncyril: **Bug (confirmed): Duplicate POST + Event on retry.**

Same pattern as the Kueue reconciler: `Eventf` fires before `promotePatch`. If `promotePatch` fails (API conflict), on retry `failureAlreadyReported=false` (annotation not persisted), `serverAlreadyHandledFailure=false` (label not persisted). Reconciler re-detects failure via `detectFailure`, re-POSTs to EvalHub, re-emits `Eventf`.

Consider: combine the promote patch (including the failure label) into one atomic update, then emit the Event.

controllers/evalhub/evaluation_job_failure_reconciler.go
```diff
@@ -328,6 +343,12 @@ func (r *EvalHubEvaluationJobFailureReconciler) Reconcile(ctx context.Context, r
 		}
 		return ctrl.Result{}, nil
 	}
+	if serverAlreadyHandledFailure(&job) {
+		log.Info("skip: EvalHub server already set failure label",
+			append(failureWatcherLogFields(), "action", "skip_server_handled",
```
sheltoncyril, line 348: **Issue (plausible): `serverAlreadyHandledFailure` skips Job deletion.**

`failureAlreadyReported` (line ~325) calls `deleteEvalHubFailureSyncedJob` before returning. This `serverAlreadyHandledFailure` path returns without cleanup. If the server sets `evaluation-phase=Failed` but does not delete the Job, it stays in the namespace indefinitely consuming quota.

Is this intentional — i.e., is the server expected to clean up the Job? If so, a comment would help. If not, consider calling `deleteEvalHubFailureSyncedJob` here too.

controllers/evalhub/evaluation_job_failure_reconciler.go
```diff
@@ -729,6 +757,16 @@ func failureAlreadyReported(job *batchv1.Job) bool {
...
 }
 
+// serverAlreadyHandledFailure returns true when the EvalHub server has already set evaluation-phase=Failed
+// on the Job, indicating the server reported the failure via its own event path. Prevents duplicate
+// Events when both the server and operator can detect the same failure.
+func serverAlreadyHandledFailure(job *batchv1.Job) bool {
+	if job.Labels == nil {
+		return false
+	}
+	return job.Labels[labelEvaluationPhase] == labelEvaluationPhaseFailed
```
sheltoncyril, line 772: **Bug (confirmed): Misleading name dedupes own operator label.**

Docstring says "EvalHub server has already set evaluation-phase=Failed" but the operator's own Kueue reconciler also sets `labelEvaluationPhase=Failed` via `patchJobFailureLabels`. This function cannot distinguish the two sources.

This is the root cause of the Kueue reconciler deadlock: on retry after a partial commit, the operator's own label triggers `serverAlreadyHandledFailure=true`, permanently blocking the Workload annotation.

Suggestion: rename to `failureLabelAlreadySet` and update callers to handle both origins explicitly.

controllers/evalhub/evaluation_job_failure_reconciler.go
```diff
@@ -409,12 +430,19 @@ func (r *EvalHubEvaluationJobFailureReconciler) Reconcile(ctx context.Context, r
...
 
+	r.EventRecorder.Eventf(&job, corev1.EventTypeWarning, eventReasonEvaluationFailed, "%s", msg)
+
 	promotePatch := client.MergeFrom(job.DeepCopy())
 	if job.Annotations == nil {
 		job.Annotations = map[string]string{}
 	}
+	if job.Labels == nil {
+		job.Labels = map[string]string{}
+	}
```
sheltoncyril, line 444: **Issue (plausible): Duplicated label/annotation patching logic.**

This inline block sets `labelEvaluationPhase` + `annotationEvaluationStatus` with fatal error handling (triggers requeue). The Kueue reconciler's `patchJobFailureLabels` method does the same with non-fatal error handling (logs and continues). If the label key or value changes, both must be updated independently.

Consider extracting a shared helper or at minimum referencing the other site with a comment.

sheltoncyril (commented):
AI code review (max depth, 10-angle sweep). 3 confirmed bugs, 9 additional findings. See inline comments.

config/components/evalhub/rbac/evalhub_events_role.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# ClusterRole for EvalHub event emission
+# Grants create on core events so the EvalHub server can emit Kubernetes
+# Events against backing Job resources on evaluation lifecycle transitions
```
sheltoncyril: **Issue (confirmed): Comment says "EvalHub server" but binding targets the operator SA.**

The `evalhub_events_binding.yaml` binds this ClusterRole to `controller-manager` (the operator SA), not the EvalHub server SA. The operator already has `events` create/patch/update from the generated manager role (via the `//+kubebuilder:rbac` markers in both reconcilers), making this static CRB potentially redundant.

At minimum, update the comment to say "operator" not "server." Or if the intent is for the server SA to emit events in tenant namespaces, this needs a runtime RoleBinding (like `createJobsWriterRoleBinding`) instead of a static CRB.

config/components/evalhub/rbac/evalhub_jobs_writer_role.yaml
```diff
@@ -11,4 +11,4 @@ metadata:
 rules:
   - apiGroups: ["batch"]
     resources: ["jobs"]
-    verbs: ["create", "delete", "list"]
+    verbs: ["create", "delete", "get", "list", "patch"]
```
sheltoncyril, line 14: **Issue (plausible): Expanded permissions broader than needed.**

`get` and `patch` on `batch/jobs` is now granted to the EvalHub server SA via `createJobsWriterRoleBinding` in every tenant namespace. This covers **all** batch Jobs in the namespace, not just EvalHub evaluation jobs (no `resourceNames` scope).

If these verbs are only needed for the operator's failure labeling (which already has its own RBAC via kubebuilder markers), consider whether the server SA truly needs them.

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
ruivieira: The described deadlock cannot happen in this code. Here's why:

1. patchJobFailureLabels runs after the serverAlreadyHandledFailure check, not before. Look at the flow in Reconcile():
    - serverAlreadyHandledFailure check → early return if true
    - postEvalHubBenchmarkFailed (POST to EvalHub server)
    - patchJobFailureLabels (sets evaluation-phase=Failed on the Job)
    - annotateWorkloadReported (annotates the Workload)
2. The scenario described: patchJobFailureLabels succeeds but annotateWorkloadReported fails, so on retry the Workload re-enters reconciliation and hits serverAlreadyHandledFailure -> early return, never annotating the Workload.

This is technically accurate as a sequence, but calling it a "deadlock" and "perpetual no-op reconcile cycles" it's incorrect. The EvalHub POST already succeeded before either patch runs. The failure event has been reported to EvalHub. The only consequence is:
    - The Workload annotation never gets written
    - The Workload keeps re-enqueuing on informer resync
    - Each time it hits serverAlreadyHandledFailure and returns immediately (cheap no-op)

So it's a minor resource waste (unnecessary reconcile cycles that exit early), not a deadlock or data-loss bug.

3. The root cause claim is wrong. Saying serverAlreadyHandledFailure "cannot distinguish the operator's own label from one set by the EvalHub server." That's true, but it's intentional: both cases mean the failure is already handled. The real issue is just that the Workload annotation should still be written even when the operator set the label itself.

@coderabbitai can you verify?

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
ruivieira: Fixed in https://github.com/trustyai-explainability/trustyai-service-operator/pull/846/commits/b51adc3659c6bf56b30b9202df498220400c5dce

@coderabbitai can you verify the fix?

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
ruivieira: Fixed in https://github.com/trustyai-explainability/trustyai-service-operator/pull/846/commits/7170627c0e98c18fe68157d3550262588ddb34c5

@coderabbitai please verify.

controllers/evalhub/evaluation_failed_kueue_workloads_reconciler.go
ruivieira: I think this is conflating Labels with Annotations. The nil-check guards `job.Annotations`, not `job.Labels`. A Job can have labels set (by EvalHub when creating it) while having no annotations at all. These are independent maps.

The Labels nil-check is indeed unreachable (since `isEvalHubEvaluationJob` already indexed into `job.Labels`), but that's a defensive pattern that's harmless and idiomatic in controller code. The Annotations nil-check is really needed.

@coderabbitai opinions?

controllers/evalhub/evaluation_job_failure_reconciler.go
ruivieira: IMO this misses the `failurePendingReport` guard that already exists in this reconciler.

This reconciler uses a two-phase commit pattern that the Kueue reconciler lacked:

  1. Before POSTing, writes annotationFailurePending = "true" on the Job. Survives retries.
  2. If the POST fails, reverts the pending annotation so the next reconcile starts fresh.
  3. Eventf fires after a successful POST.
  4. promotePatch atomically swaps failure-pending → failure-reported + sets the label.

If `promotePatch` fails on retry:

- `failurePendingReport(&job)` returns true, so it skips the pending-annotation write and goes straight to the POST
- The POST fires again: this is the duplicate the reviewer flags

**But**:

- The failure-pending annotation acts as a dedup gate preventing the full detection path from re-running
- The POST is naturally idempotent by Job ID (it reports the same failure for the same job)
- Event duplication is mitigated by K8s event aggregation (as the reviewer notes)

@coderabbitai wdyt?

controllers/evalhub/evaluation_job_failure_reconciler.go
ruivieira: Fixed in https://github.com/trustyai-explainability/trustyai-service-operator/pull/846/commits/87bb77f71d5d0fe31e6a83d796878be430cd49ed

@coderabbitai please verify the fix

controllers/evalhub/evaluation_job_failure_reconciler.go
ruivieira: Already fixed it in b51adc36

controllers/evalhub/evaluation_job_failure_reconciler.go
ruivieira: I think this is comparing two different things:

* Job failure reconciler: The promotePatch block does more than just set labels (it atomically swaps failure-pending -> failure-reported), sets the evaluation status annotation, and sets the phase label, all in one patch. This is part of the two-phase commit protocol specific to this reconciler.
* Kueue reconciler's patchJobFailureLabels: A standalone method that only patches the label and status annotation on a Job it doesn't own the lifecycle of

These have different error handling logic (fatal vs requeue), different state transitions, and different transactional boundaries. Extracting a shared helper would force one to fit the other (the Job failure reconciler would lose the atomic failure-pending deletion, or the helper would need conditional logic for annotations it may or may not manage).

@coderabbitai opinions?

config/components/evalhub/rbac/evalhub_events_role.yaml
ruivieira: Fixed comment in https://github.com/trustyai-explainability/trustyai-service-operator/pull/846/commits/c3440b1b97d60cdb7e6af2b0baa78470b10b9e2d

config/components/evalhub/rbac/evalhub_jobs_writer_role.yaml
ruivieira: The get and patch verbs were added in preparation for RHAI-277 that has as acceptance criteria "EvalHub ServiceAccount has RBAC permissions to create Events and patch Jobs"

controllers/evalhub/evaluation_job_failure_reconciler.go
ruivieira: Fixed in https://github.com/trustyai-explainability/trustyai-service-operator/pull/846/commits/9bdbfe3bfd199a784313e138614e616fd4249af6

@coderabbitai verify, please

sheltoncyril (approved):

PR #859, opened 2026-08-08
feat(evalhub): instrument EvalHubReconciler.Reconcile() with metrics

ruivieira:
Use named returns + a single defer to record reconcile duration,
result label (success/requeue/error), and error type on every return
path without modifying individual return sites. Covers all 10+
return paths including handleDeletion and the periodic ready-state
requeue ([RHAI-240](https://redhat.atlassian.net/browse/RHAI-240)).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


julpayne (approved):
LGTM

ppadashe-psp (approved):
Approved

PR #860, opened 2026-08-08
feat(evalhub): update managed instances gauge after each full reconcile cycle

ruivieira:
Add countManagedInstances() helper that lists all EvalHub CRs cluster-wide
and call setManagedInstances() after updateStatus() succeeds, so the gauge
reflects the live CR count on both the ready and not-yet-ready requeue paths
while skipping early-exit error paths ([RHAI-240](https://redhat.atlassian.net/browse/RHAI-240)).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/controllers/evalhub/evalhub_controller.go
+++ b/controllers/evalhub/evalhub_controller.go
@@ -334,12 +334,22 @@ func (r *EvalHubReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		return RequeueWithError(err)
 	}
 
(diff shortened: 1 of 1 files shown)


ppadashe-psp:
@ruivieira  @julpayne  Can we get this reviewed and merged as this is targeted for 3.6EA2 ?

PR #877, opened 2026-08-14
feat(evalhub): add EvalHub controller distributed tracing (RHAI-241)

julpayne:
## Summary

- Add opt-in OpenTelemetry tracing for the EvalHub operator controller via standard `OTEL_*` environment variables (no manifest or CLI changes).
- Introduce `pkg/tracing` bootstrap with noop provider when OTLP is not configured, and instrument `EvalHubReconciler` with parent/child reconcile spans (`evalhub.reconcile` + phase children per [RHAISTRAT-1606](https://redhat.atlassian.net/browse/RHAISTRAT-1606)).
- Instrument `EvalHubEvaluationJobFailureReconciler` with `evalhub.job_failure_reconcile` spans and document operator vs workload OTEL in `controllers/evalhub/OTEL_TRACING.md`.

Closes [RHAI-241](https://redhat.atlassian.net/browse/RHAI-241).

## Changes to `cmd/main.go`

This PR wires operator-wide tracing bootstrap into the manager entrypoint and adjusts startup flow so tracing can shut down cleanly on every exit path.

### OpenTelemetry bootstrap

After logger setup and flag parsing, `run()` calls `tracing.Setup(context.Background())` from `pkg/tracing`. That function:

- Installs a **noop** `TracerProvider` when OTLP is not configured (no endpoint env vars, or `OTEL_SDK_DISABLED=true`), so there is no runtime overhead when tracing is off.
- Creates an OTLP trace exporter and batching `TracerProvider` when `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is set.

On success, `run()` registers a deferred shutdown callback:

```go
defer func() {
    if err := traceShutdown(context.Background()); err != nil {
        setupLog.Error(err, "problem shutting down OpenTelemetry tracing")
    }
}()
```

This flushes and shuts down the SDK exporter when the process exits normally or after a startup failure.

### `main()` / `run()` refactor

Startup logic was moved from `main()` into `run() int`, with `main()` reduced to:

```go
func main() {
    os.Exit(run())
}
```

All previous `os.Exit(1)` calls inside startup (missing services, TLS resolution failure, manager/webhook/controller setup errors, health check registration, manager start failure) now `return 1` instead.

**Why:** Calling `os.Exit` directly skips deferred functions in the current goroutine. Returning an exit code to `main()` ensures the `traceShutdown` defer runs on every failure path after tracing has been initialized, while preserving existing log messages and exit semantics.

Tracing setup failure itself returns `1` before the defer is registered (no provider was installed), matching prior behavior.

### Operational impact

- No new CLI flags or Deployment manifest changes are required for tracing; configuration remains env-var driven (`OTEL_*`).
- When OTLP is not configured, operator behavior is unchanged aside from the noop provider registration at startup.
- Tracing is process-global: the bootstrap in `cmd/main.go` enables any controller using `pkg/tracing`, not only EvalHub (though only EvalHub is instrumented in this PR).

## Test plan

- [x] `go test ./pkg/tracing/... ./controllers/evalhub/... -run 'Tracing|TestEvalHubReconcile|TestExitCode|TestTruncate|TestJobFailure|TestSetup|TestWithPhase|TestSetReconcile'`
- [ ] Deploy operator with `OTEL_EXPORTER_OTLP_ENDPOINT` set and confirm spans appear in OTLP backend
- [ ] Confirm noop behavior when OTLP env vars are unset

## Summary by CodeRabbit

* **New Features**
  * Added opt-in OpenTelemetry tracing for application startup and EvalHub reconciliation.
  * Added support for tracing reconciliation phases, failures, actions, outcomes, and relevant job or pod details.
  * Supports OTLP exporters over gRPC or HTTP, service-name overrides, and configurable tracing behavior.

* **Documentation**
  * Added setup, configuration, deployment, verification, and troubleshooting guidance for tracing.

* **Tests**
  * Added coverage for tracing setup, span attributes, failure reporting, outcomes, and exit-code detection.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
@@ -119,7 +120,9 @@ func run() int {
 		return 1
 	}
 	defer func() {
-		if err := traceShutdown(context.Background()); err != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := traceShutdown(ctx); err != nil {
 			setupLog.Error(err, "problem shutting down OpenTelemetry tracing")
 		}
 	}()
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -89,6 +89,10 @@ func init() {
 }
 
 func main() {
+	os.Exit(run())
+}
+
+func run() int {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
@@ -112,7 +116,7 @@ func main() {
 	traceShutdown, err := tracing.Setup(context.Background())
 	if err != nil {
 		setupLog.Error(err, "unable to set up OpenTelemetry tracing")
-		os.Exit(1)
+		return 1
 	}
 	defer func() {
 		if err := traceShutdown(context.Background()); err != nil {
@@ -122,14 +126,14 @@ func main() {
 
 	if enabledServices.Empty() {
 		setupLog.Error(fmt.Errorf("no service is specified"), "please specify at least one service")
-		os.Exit(1)
+		return 1
 	}
 
 	cfg := ctrl.GetConfigOrDie()
 	tlsResult, err := pkgtls.Resolve(context.Background(), cfg)
 	if err != nil {
 		setupLog.Error(err, "unable to resolve TLS configuration")
-		os.Exit(1)
+		return 1
 	}
 	tlsOpts := tlsResult.TLSOpts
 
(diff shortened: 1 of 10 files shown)


cmd/main.go
```diff
@@ -108,16 +113,27 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	traceShutdown, err := tracing.Setup(context.Background())
+	if err != nil {
+		setupLog.Error(err, "unable to set up OpenTelemetry tracing")
+		return 1
+	}
+	defer func() {
```
ruivieira, line 122: @julpayne Would it make sense to use a timeout-bounded context for the OTEL shutdown, rather than context.Background()? If the exporter's gRPC connection is stale the shutdown call can block until the OTLP timeout, blocking pod termination by that amount silently.

Instead of

```go
    defer func() {
        if err := traceShutdown(context.Background()); err != nil {  // no deadline
```

Having something like

```go
    _ = tracing.WithPhase(ctx, spanReconcileRoute, func(ctx context.Context) error {
        if err := r.reconcileRoute(ctx, instance); err != nil {
            log.Error(err, "Failed to reconcile Route")
        }
        return nil   // error swallowed, span always succeeds
    })
```

ruivieira (commented):
LGTM, thanks @julpayne 

Just a minor comment

cmd/main.go
julpayne: @ruivieira I have fixed this - thanks P.S. I think that your suggestion was a bad paste?

ruivieira (approved):

PR #891, opened 2026-08-27
fix(tas): authenticate Prometheus scraping through kube-rbac-proxy

SudipSinha:
## Summary

Fixes Prometheus metrics scraping by authenticating *through* kube-rbac-proxy instead of bypassing it (supersedes #885, which raised security concerns about direct port exposure).

## Root cause

Two issues prevented Prometheus from scraping TrustyAI metrics:

1. **Wrong port**: ServiceMonitor targeted `http`/8080 (health-only app), which does not serve `/q/metrics`
2. **No auth token**: Empty `bearerTokenSecret: key: ''` prevented Prometheus from presenting any token, so kube-rbac-proxy rejected every scrape

## Changes

- **Dedicated metrics-reader ServiceAccount** per TrustyAI namespace with a long-lived token Secret — avoids coupling to `prometheus-user-workload` SA which may or may not exist
- **ServiceMonitor** targets kube-rbac-proxy (`https`/8443), `scheme: https`, `insecureSkipVerify: true`, with `bearerTokenSecret` referencing our token
- **Role** grants `get services` (kube-rbac-proxy's SAR check) and `get secrets/<token>` (so Prometheus SA can read the token to mount it)
- **RoleBinding** binds both our metrics-reader SA (for kube-rbac-proxy auth) and `prometheus-user-workload` SA (to read the token Secret)
- **CA bundle ConfigMap** created with service-CA injection annotation (available for future use)
- **Selector tightened** with `trustyai-service-name` label to match only the `-tls` service
- **Drift reconciliation** for ServiceMonitor specs and RBAC on upgrade
- **kubebuilder markers** for `roles`/`rolebindings`
- **Kubernetes events** recorded on resource creation (per coding guidelines)
- **`controllerutil.SetControllerReference`** used throughout (per coding guidelines)

## Test plan

- [x] All 45/45 Ginkgo specs pass
- [x] `go build ./controllers/tas/...` succeeds
- [x] On-cluster: Prometheus successfully scrapes `trustyai_*` metrics through kube-rbac-proxy (verified: target `health: up`, HTTP 200 from kube-rbac-proxy)

## Summary by CodeRabbit

* **New Features**
  * Metrics monitoring now uses secure HTTPS endpoints with certificate validation.
  * Added automatic certificate bundle configuration for metrics collection.
  * Added authenticated metrics access using a dedicated service account and token.
  * Added permissions that allow Prometheus to securely retrieve service metrics.
  * Added status events for newly created monitoring resources.

* **Bug Fixes**
  * Existing monitoring configurations are now updated when their settings change.
  * Corrected an inaccurate monitoring error message.
  * Fixed metrics authorization configuration to correctly identify services.


controllers/tas/monitor.go
```diff
@@ -120,6 +134,224 @@ func (r *TrustyAIServiceReconciler) ensureLocalServiceMonitor(cr *trustyaiopenda
...
+			Labels: map[string]string{
+				"app.kubernetes.io/part-of": componentName,
+			},
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"services"},
+				Verbs:     []string{"get"},
+			},
```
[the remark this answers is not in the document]
SudipSinha: Addressed in the latest commit (`e31810e`): `resourceNames: [cr.Name]` is restored now that the kube-rbac-proxy config correctly uses `name:` instead of `resourceName:`. The previous commit dropped the restriction as a workaround because the SAR was omitting the resource name (due to the wrong config field), making the scoped Role always deny. With the config fixed, the SAR correctly includes the name and the restriction works as intended.

controllers/tas/templates/service/service-monitor-local.tmpl.yaml
```diff
@@ -21,11 +22,14 @@ spec:
                   - '{__name__= "trustyai_spd"}'
                   - '{__name__= "trustyai_dir"}'
           path: /q/metrics
-          port: http
-          scheme: http
+          port: https
+          scheme: https
+          tlsConfig:
+              insecureSkipVerify: true
```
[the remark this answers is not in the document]
SudipSinha: Deliberate tradeoff. The CA bundle ConfigMap is populated asynchronously by the OpenShift service-CA operator after the ConfigMap is created. Using CA verification caused Prometheus scrapes to fail in the test window before the injection completed. Since the bearer token already provides authentication through kube-rbac-proxy, `insecureSkipVerify: true` means the channel is encrypted but the server cert is not verified — an acceptable risk on an internal cluster network where the traffic never leaves the cluster. Proper CA verification can be restored in a follow-up once the async injection timing is handled (e.g. waiting for the ConfigMap to be populated before creating the ServiceMonitor).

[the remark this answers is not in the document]
SudipSinha: Correction to my previous reply: the original ServiceMonitor used `scheme: http` with no encryption at all. Switching to `scheme: https` with `insecureSkipVerify: true` is a net security improvement — traffic is now encrypted and bearer token auth is enforced through kube-rbac-proxy. The only gap vs full TLS is server cert verification, which is acceptable on an internal cluster network where traffic never leaves the cluster. Proper CA verification can be added as a follow-up once the async injection timing of the CA bundle ConfigMap is handled.

m-misiura:
```
      tlsConfig:
              insecureSkipVerify: true
```
is this a problem?

controllers/tas/monitor.go
```diff
@@ -120,6 +130,189 @@ func (r *TrustyAIServiceReconciler) ensureLocalServiceMonitor(cr *trustyaiopenda
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      saName,
+				Namespace: cr.Namespace,
+			},
+			{
+				Kind:      "ServiceAccount",
+				Name:      "prometheus-user-workload",
```
m-misiura, line 317: is this hardcoding OK?

controllers/tas/trustyaiservice_controller.go
```diff
@@ -237,6 +239,18 @@ func (r *TrustyAIServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return RequeueWithError(err)
 	}
 
+	// Metrics reader ServiceAccount + token (must exist before ServiceMonitor references it)
+	err = r.ensureMetricsReaderServiceAccount(instance, ctx)
```
m-misiura, line 243: the rest of the reconciler uses RequeueWithError(err)

controllers/tas/templates/service/service-monitor-local.tmpl.yaml
SudipSinha: Yes it was — but it turned out to be unnecessary. We introduced it as a workaround believing the OpenShift service-CA operator injects the CA bundle asynchronously after ConfigMap creation, which we thought would cause Prometheus scrape failures. We tested this manually and the CA bundle was already fully populated by the time the TrustyAI pod reached Running state. `curl --cacert <ca-bundle>` with a valid bearer token returned 200. `insecureSkipVerify: true` has been reverted in the latest commit — proper CA + serverName verification is now restored.

controllers/tas/monitor.go
SudipSinha: Addressed in commit `6229751`: extracted to `prometheusUserWorkloadNamespace` constant in `constants.go`.

controllers/tas/trustyaiservice_controller.go
SudipSinha: Fixed in commit `6229751` — all new reconciliation steps now use `RequeueWithError(err)` consistently with the rest of the controller.

m-misiura (approved):

PR #875, opened 2026-08-13
test(module): Add platform contract conformance tests for TrustyAI type

kpunwatk:
Related to: https://redhat.atlassian.net/browse/RHOAIENG-83575
Adds unit tests validating TrustyAI implements the PlatformObject contract (status, conditions, release status, singleton naming) via odh-platform-utilities' validation helper. Defines the missing TrustyAIInstanceName constant to match the existing CEL singleton validation (metadata.name == 'default').

## Summary by CodeRabbit

* **New Features**
  * Established `default-trustyai` as the supported name for TrustyAI instances, providing consistent instance identification.

* **Tests**
  * Expanded validation coverage for platform compatibility, status handling, condition and release-status persistence, and singleton instance-name rules.

PR #893, opened 2026-08-31
Feat: module envtest scaffolding

kpunwatk:
Summary                                                                                                                                                                       
                                                                                                                                                                                
  Stands up envtest/Ginkgo test scaffolding for trustyai-operator-module, which had zero test coverage outside the type-level conformance test in #875. Part of the ongoing work
  to close the E2E testing-responsibility sign-off gap (module operator side of [RHOAIENG-79650](https://redhat.atlassian.net/browse/RHOAIENG-79650)).                                                                                
                                                                                                                                                                                
  What's included                                                                                                                                                               
                                                                                                                                                                                
  Manifest fix: config/crd/bases/components.platform.opendatahub.io_trustyais.yaml and config/rbac/role.yaml had not been regenerated since #854 switched conditions to         
  odh-platform-utilities' common.Condition. The stale CRD still required non-empty reason/message on every condition (leftover from the old metav1.Condition-shaped schema).    
  This causes status updates to fail with a 422 whenever a condition is seeded without a reason — e.g. handleRemoval() when ManagementState is set to Removed before the        
  module's first full reconcile. Found this via the new tests below; fixed by regenerating both manifests from current markers/types.                                           
                                                                                                                                                                                
  Test scaffolding:                                                                                                                                                             
  - pkg/trustyaimodule/suite_test.go — envtest bootstrap, following the controllers/*/suite_test.go pattern used elsewhere in this repo.                                        
  - pkg/trustyaimodule/reconciler_test.go — covers:                                                                                                                             
    - Singleton CEL enforcement (accepts default, rejects any other name)                                                                                                       
    - Finalizer added on first reconcile                                                                                                                                        
    - observedGeneration tracking after a successful reconcile                                                                                                                  
    - DSC ConfigMap + finalizer cleanup on deletion                                                                                                                             
    - Ready/Degraded condition transitions when ManagementState: Removed
    
   Scope / what's deferred                                                                                                                                                       
                                                                                                                                                                                
  Reconciler tests target the current pre-#872 reconciler.go. Health-status transitions, garbage collection, and dependency-check (KServe/Prometheus) coverage are deferred     
  until #872 (reconciler/health/dependencies rewrite) lands, since writing that coverage against the current code would be discarded. Tracked in                                
  test-coverage-followup-ticket.md.                                                                                                                                             
                                                                                                                                                                                
  Also noted but not fixed here: handleRemoval doesn't backfill the DependenciesAvailable/KServeAvailable conditions, so they're left in a seeded "Unknown, no reason" state if 
  Removed is set before the first full reconcile. This is schema-legal now (reason/message are optional) but still semantically sloppy — left as-is since reconciler.go is one  
  of the files #872 rewrites.                                                                                                                                                   
                                                                                                                                                                                
  Test plan                                                                                                                                                                     
                                                                                                                                                                                
  - [x] go build ./..., go vet ./... clean                                                                                                                                      
  - [x] go test ./... -v — 6/6 specs pass                                                                                                                                       
  - [x] gofmt -l clean on new files
    

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated TrustyAI module status and release metadata schemas for improved compatibility and validation.
  - TrustyAI module resources must now use the `default-trustyai` name.
  - Refined permissions for managed Kubernetes resources.

- **Tests**
  - Added reconciliation coverage for validation, finalizers, generation updates, cleanup, and removed management state.
  - Added dependency checks for Prometheus availability and controller behavior.
  - Added an integration test environment and supporting Prometheus resource definitions.

- **Documentation**
  - Improved descriptions for management state, conditions, phases, and release information.
  - Clarified supported management states and release metadata requirements.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


trustyai-operator-module/pkg/trustyaimodule/suite_test.go
```diff
@@ -0,0 +1,73 @@
...
+		ErrorIfCRDPathMissing: true,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// This reuses the envtest binaries downloaded for the main operator module via
+		// `make envtest` at the repo root, since both modules test against the same
+		// Kubernetes version.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
+			fmt.Sprintf("1.29.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
ruivieira: @kpunwatk should we read this from the env var `ENVTEST_K8S_VERSION` and default to 1.29.0 if missing? If the Makefile bumps the version this would fail.

ruivieira (changes requested):

trustyai-operator-module/pkg/trustyaimodule/suite_test.go
kpunwatk: @ruivieira I've addressed and pushed the changes.

While doing this change I noticed this exact pattern (hardcoded version + comment) is the standard kubebuilder-scaffolded boilerplate already used in every other suite_test.go in this repo (e.g. controllers/nemo_guardrails/suite_test.go). Also, BinaryAssetsDirectory is overridden by KUBEBUILDER_ASSETS per controller-runtime's own docs, and the Makefile sets that env var from the   
current ENVTEST_K8S_VERSION when running via make test-tom. So a Makefile bump wouldn't break the normal test run — only a direct go test invocation that bypasses the Makefile.

ruivieira: @kpunwatk thanks. My original comment was about this flow:

* `ENVTEST_K8S_VERSION` is bumped on the makefile from (say) 1.29.0 to 1.30.0
* `make test` downloads binaries to bin/k8s/1.30.0-linux-amd64/ and runs fine
* Someone runs `go test ./controllers/...` directly
* envtest falls back to BinaryAssetsDirectory, which still points to bin/k8s/1.29.0-linux-amd64/

We can make the other tests more robust too (but in future PRs)

ruivieira (approved):

PR #895, opened 2026-09-01
Feat: Add ability to route usernames to a specific guardrails configuration

RobGeada:
Adds `spec.nemoConfigs.usernames` and a user-mapping-proxy layer to NeMo Guardrails, that can automatically enforce that users or user groups will be mandatorily routed to specified guardrails configurations.

```
kind: NemoGuardrails
metadata:
  name: nemoguardrails
  annotations:
    security.opendatahub.io/enable-auth: 'true'
spec:
  nemoConfigs:
    - name: main-config
      configMaps:
        - nemo-config-main
      default: true
      usernames:
        - "email@email.com"
        - "system:serviceaccount:model-namespace:nemo-guardrails-usergroup-alpha"  # to use a servicea ccount 
        - "nemo-guardrails-usergroup-alpha" # the service account short-name is also accepted - it will enforce that the namespace matches
    - name: nemo-config-second
      configMaps:
        - nemo-config-second
      usernames:
        - "user-d@email.com"
        - "nemo-guardrails-usergroup-beta" 
  env:
    - name: "OPENAI_API_KEY"
      value: "null"
```

## Architecture
1) The TrustyAI Service operator creates a configmap based on the provided `nemoConfigs[*].usernames` mapping.
2) Kube-rbac-proxy: Authenticates the request, adds the `x-remote-user` header
3) user-mapping-proxy: Compares the `x-remote-user` header against the mounted user-mapping configmap. If a username is matched to a config, then the `guardrails.config_id` field is set/overwritten in their request. If the username does not match any config, then the `guardrails.config_id` is deleted to enforce fallback routing to the default guardrails config.

## Summary by CodeRabbit

- **New Features**
  - Added optional username-based routing for Nemo Guardrails configurations.
  - Authenticated users can be routed to the first configuration matching their username.
  - Users without a matching username continue to use the default configuration.
  - Added support for authentication-aware routing through the user-mapping proxy.

- **Configuration**
  - Added the `usernames` field to Nemo Guardrails configuration.
  - Username-based routing requires authentication to be enabled.


christinaexyou:
@RobGeada i think a new CRD, something like`NemoGuardrailsUserRouting`, would be more appropriate to orchestrate the logic here. by separating the routing policy from the service, we 1) better prevent the NemoGuardrails controller from reconciling when `usernames` are updated 2) errors related to the routing policy are less likely to break the service

config/base/kustomization.yaml
```diff
@@ -149,3 +149,10 @@ vars:
       apiVersion: v1
     fieldref:
       fieldpath: data.nemo-guardrails-image
+  - name: nemo-guardrails-reverse-proxy-image
+    objref:
+      kind: ConfigMap
```
christinaexyou, line 154: i'm not sure if a ConfigMap is the right resource to store the user routing policy; reason being that anyone with ConfigMap read access can read usernames

controllers/nemo_guardrails/nemoguardrail_controller.go
```diff
@@ -137,12 +137,46 @@ func (r *NemoGuardrailsReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	// ====== Deploy user-mapping configmap if needed ===================================================================
+	var userMappingCM *corev1.ConfigMap
+	if utils.RequiresAuth(nemoGuardrails) {
+		configMapping := make(map[string][]string)
```
christinaexyou, line 143: i would double check that Go hashmaps are deterministic. we document that routing always defaults to the first matching but what if a username appears in multiple configs

PR #896, opened 2026-09-01
feat(tls): add SecureServing metrics and TLS profile watcher

ugiordan:
## Summary

- **SecureServing metrics**: switch from plain HTTP `:8080` to HTTPS `:8443` with controller-runtime `SecureServing` and `FilterProvider` authentication (SubjectAccessReview)
- **ProfileWatcher**: monitors `apiservers.config.openshift.io/cluster` and cancels the manager context on TLS profile change, triggering a pod restart
- **Forbidden/Unauthorized handling**: if the operator lacks RBAC to read the TLS profile at startup, it still registers the watcher (`APIAvailable=true`) so it retries automatically when RBAC is fixed
- **RBAC overlays**: metrics-auth `ClusterRole`/`ClusterRoleBinding` for `odh` and `rhoai` overlays
- **OpenShift TLS wiring**: serving-cert Secret mount, Service annotation, and ServiceMonitor service-CA TLS in both overlays
- Register `configv1` types on the manager scheme so `ProfileWatcher` can watch `APIServer` resources

## Test plan

- [x] `go test ./pkg/tls/...`
- [x] `go build ./...`
- [x] `kustomize build config/overlays/odh` succeeds
- [ ] Deploy on OpenShift with rhoai overlay, verify metrics endpoint is HTTPS on port 8443
- [ ] Verify `ServiceMonitor` scrapes successfully with service-ca TLS
- [ ] Verify TLS profile change triggers pod restart

## Related

- [RHOAIENG-72826](https://redhat.atlassian.net/browse/RHOAIENG-72826), [RHOAIENG-75336](https://redhat.atlassian.net/browse/RHOAIENG-75336)

## Summary by CodeRabbit

- **New Features**
  - Metrics are now served securely over HTTPS by default on port 8443.
  - Added configurable TLS certificate settings for the metrics endpoint.
  - Metrics access now supports authentication and authorization.
  - OpenShift deployments automatically configure TLS certificates, secure scraping, and service-account permissions.
  - The operator responds to cluster TLS security profile changes by restarting with updated settings.

- **Bug Fixes**
  - Improved handling of unavailable or restricted TLS configuration APIs while retaining secure defaults.


cmd/main.go
```diff
@@ -215,7 +240,20 @@ func run() int {
...
+	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
+	defer cancel()
+
+	if tlsResult.APIAvailable {
+		watcher := pkgtls.NewProfileWatcher(mgr.GetClient(), tlsResult.ProfileSpec, func() {
+			setupLog.Info("TLS security profile changed, shutting down for restart")
+			cancel()
+		})
+		if err := watcher.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "Failed to set up TLS security profile watcher; profile changes will not trigger a restart")
```
ruivieira, line 252: @ugiordan IIUC not failing here is intentional. But if the watcher fails to register and the TLS profile changes, would this be using stale TLS profiles? Just confirming if you think this is ok.

ruivieira (approved):
@ugiordan LGTM, just left a question. Thanks!

cmd/main.go
ugiordan: Nope, that long-and-continue wasn't intentional for compliance; it's inconsistent with how I've been implementing this in odh-operator and mlflow, where we fail closed. Indeed, with this approach, if watcher registration fails, we'd keep the startup profile and miss changes until an unrelated restart.

Pushed the proper fix with an exit non-zero error if `SetupWithManager` fails, so kubelet restarts until the watcher registers (or RBAC/scheme is fixed).

ruivieira (approved):

PR #902, opened 2026-09-07
maintenance: Delete FMS Guardrails ecosystem from operator

RobGeada:
## Summary by CodeRabbit

* **Removed Capabilities**
  * Removed support for the Guardrails Orchestrator service, including its custom resource, controller, deployment, routes, monitoring, authentication, and automatic configuration.
  * Removed Guardrails-related image settings for the orchestrator, built-in detectors, and sidecar gateway.
  * The TrustyAI module no longer exposes or enables the Guardrails Orchestrator service.

* **Documentation**
  * Updated architecture, component, and supported-services documentation to exclude Guardrails Orchestrator.

* **Configuration**
  * Removed associated Kubernetes manifests, permissions, policies, and overlay configuration.


sheltoncyril:
## Regression testing results

Ran a regression pass against a build of this PR (image built from `667acced`, `linux/amd64`) on an OpenShift AI cluster, to check that removing the FMS Guardrails ecosystem doesn't cause collateral damage to the other operator-managed components.

### What was validated ✅
- **Build & startup:** image builds and the controller starts; all remaining controllers reach "Starting workers" (`TAS`, `LMES`, `NEMO_GUARDRAILS`, `EVALHUB`).
- **FMS/GORCH removal is clean:** the operator correctly rejects `GORCH` in `--enable-services` (`service GORCH is not supported`), confirming the ecosystem is gone as intended.
- **No regression to other components:**
  - NeMo Guardrails tier1: 10 passed (core config-update, LLM-as-judge, secret-mounting all green).
  - LM-Eval smoke: passed.
  - TrustyAI Service / EvalHub controllers reconcile normally.

### Notes / out of scope
- Testing was done via an **image-only hot-swap** onto an existing released install, not a full bundle/CSV install. Two RBAC gaps surfaced that stem from running a newer manager against the older released bundle's RBAC, **not from this PR's diff**:
  - `tls-profile-watcher` needs `list/watch` on `apiservers.config.openshift.io` (cache-sync is fatal without it).
  - EvalHub reconcile depends on the runtime `evalhub-events` RoleBinding.
  Both are expected to be covered when the operator is installed through its bundle/CSV; flagging only so the hot-swap caveat is explicit.
- MCP-gateway tests were skipped/errored due to ServiceMesh 3 + mcp-gateway operators not being installed on the test cluster (environment, unrelated to this PR).

### Bottom line
From a functional-regression standpoint, the FMS Guardrails deletion is clean and does not break the other components. The only follow-up on the test side is unrelated to this operator change: the downstream test suite still references `guardrailsorchestrator`, which will be handled separately.

sheltoncyril (approved):

PR #923, opened 2026-09-14
feat(module): support MCP Guardrails mode

ruivieira:
## Summary

Add support for `mcpGuardrailsMode` in the TrustyAI module operator.

When MCP Guardrails mode is enabled, the module operator deploys and configures the TrustyAI workload operator to run only the NeMo Guardrails service:

Existing resources omitted from the MCP overlay are intentionally not deleted during the mode transition.

## Motivation

The OpenDataHub DataScienceCluster supports enabling MCP Guardrails mode through:

```yaml
  spec:
    components:
      trustyai:
        managementState: Managed
        mcpGuardrailsMode: true
```

The setting is propagated to the TrustyAI module resource as:

```yaml
  spec:
    mcpGuardrailsMode: true
```

The TrustyAI module operator previously had no corresponding API field or reconciliation behavior. This change makes the module operator understand the flag and select the appropriate NeMo-only configuration.

## Behavior

When mcpGuardrailsMode is enabled:

1. The module operator selects the MCP Guardrails Kustomize overlay.
2. The workload operator Deployment is configured with:

```
     --enable-services=NEMO_GUARDRAILS
```

3. Only the NeMo Guardrails controller is enabled.
4. Module health checks wait only for the NeMo Guardrails service.
5. MCP mode takes precedence over any enabledServices values.
6. The operator Deployment is updated and restarted through normal reconciliation.

When mcpGuardrailsMode is disabled, normal ODH/RHOAI overlay selection and service configuration continue to apply.

## Resource lifecycle

This change deliberately does not delete resources that are omitted from the selected overlay during a mode switch.

For example, switching an existing deployment to MCP mode may leave previously created non-NeMo resources such as:

- non-NeMo CRDs;
- RBAC objects;
- Services;
- ServiceMonitors;
- existing operand resources.

Those resources are not actively reconciled by the NeMo-only workload operator, but they are retained. Full cleanup continues to use the existing module removal/deletion lifecycle.

This avoids introducing destructive deletion behavior into a mode switch.

### Module API and CRD

- Add spec.mcpGuardrailsMode to TrustyAISpec.
- Update the TrustyAI module CRD schema.
- Document that MCP mode takes precedence over enabledServices.

### Overlay selection

- Add MCP-aware overlay selection to EnsureManifests.
- MCP mode selects:

   overlays/mcp-guardrails

regardless of whether the platform is ODH or RHOAI.

- Normal platform-based overlay selection remains unchanged.
- Manifest copying still happens once per process.
- Overlay selection and image parameter injection occur on each reconciliation, so changing the mode on an existing CR causes the next reconciliation to render the new configuration without restarting the module operator.

  ### Service selection

- Add MCP-aware effective service selection.
- When MCP mode is enabled, the effective service set is:

    NEMO_GUARDRAILS

- Any conflicting enabledServices values are ignored while MCP mode is active.
- Normal empty-service behavior continues to enable all TrustyAI services outside MCP mode.

### Cleanup behavior

- Cluster-scoped RBAC cleanup during module deletion continues to render the normal platform overlay.
- No additional deletion is performed when switching between normal and MCP mode.

## Tests

Added unit coverage for:

- MCP mode forcing NeMo Guardrails only;
- normal service selection remaining unchanged outside MCP mode;
- MCP overlay selection taking precedence over platform type;
- normal ODH/RHOAI overlay selection when MCP mode is disabled;
- existing service argument injection behavior.

## Verification

The TrustyAI module packages compile successfully.
Both the normal and MCP Guardrails Kustomize overlays render successfully.

## Compatibility

- Existing TrustyAI deployments remain unchanged when mcpGuardrailsMode is false or omitted.
- MCP mode is authoritative over enabledServices.
- Mode changes update the workload operator configuration but do not delete resources omitted from the selected overlay.
- No OpenDataHub operator changes are included in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added an optional MCP Guardrails mode for TrustyAI deployments.
  * When enabled, deployments include only the NeMo Guardrails service and take precedence over other enabled-service settings.
  * Changes to the selected mode are applied during the next reconciliation.

* **Bug Fixes**
  * Improved cleanup to ensure all previously deployed resources are handled, including resources outside the MCP-only configuration.


trustyai-operator-module/pkg/trustyaimodule/manifests.go
```diff
@@ -31,44 +31,43 @@ func manifestsTarget() string {
...
-		return "", fmt.Errorf("clearing manifests target %s: %w", dst, err)
-	}
-	if err := copyDir(src, dst); err != nil {
-		return "", fmt.Errorf("copying manifests from %s to %s: %w", src, dst, err)
+	if copyErr != nil {
+		return "", copyErr
 	}
 
-	overlay := selectOverlay(dst)
+	overlay := selectOverlay(manifestsTarget(), mcpMode)
```
christinaexyou: nit: this is v low probability but iiuc if we were to have multiple controller runtimes, we could have 2 reconcilers write to params.env. this doesn't need to be addressed if it's not a real risk

trustyai-operator-module/pkg/trustyaimodule/reconciler.go
```diff
@@ -437,7 +420,9 @@ func (r *TrustyAIModuleReconciler) updateReleases(module *platformv1alpha1.Trust
 }
 
 // reconcileComponent renders the Kustomize overlay for the trustyai-service-operator
-// and SSA-applies all resources into the cluster. On failure it marks
+// and SSA-applies all resources into the cluster. Overlay selection changes are
+// applied on the next reconciliation; resources omitted by the selected
+// overlay are not deleted during a mode switch. On failure it marks
```
christinaexyou, line 425: should we log a message in the controller during a mode switch from non-MCP -> MCP mode ?

trustyai-operator-module/pkg/trustyaimodule/manifests.go
ruivieira: Fixed in https://github.com/trustyai-explainability/trustyai-service-operator/pull/923/commits/affbe67458be7bb5bc6988a6bc6c9e18092fe4bd

@coderabbitai does it address the comment?

trustyai-operator-module/pkg/trustyaimodule/reconciler.go
ruivieira: @christinaexyou although I agree that would be a good addition, I think it would imply having some `status.observedMcpGuardrailsMode` or similar field in order to know the previous mode. We could perhaps leave it to a separate PR?

christinaexyou (approved):

PR #928, opened 2026-09-16
fix(module): report ready without operand instances

ruivieira:
## Summary

Update the TrustyAI module readiness logic so that optional operand instances do not prevent the TrustyAI module from becoming ready, while still verifying that the TrustyAI workload operator itself is healthy.

## Changes

- Treat a service with no operand instances as healthy.
- Add a health check for the TrustyAI service operator Deployment.
- Verify the operator Deployment:
  - Exists.
  - Has observed the current generation.
  - Has updated, ready, and available replicas.
  - Does not have a replica failure condition.
- Preserve unhealthy status for existing operands that fail readiness checks.
- Update unit and reconciler tests for the new readiness behavior.

## Readiness scenarios

| Operator Deployment | Operand instances | Result |
|---|---|---|
| Not found or not ready | None | `Ready=False` |
| Ready | None | `Ready=True` |
| Ready | All operands healthy | `Ready=True` |
| Ready | One or more operands unhealthy | `Ready=False` |
| Ready | Operand list cannot be read | `Ready=Unknown` |

No operand instances is now reported as:

```text
No operand instances found
```

This means that operand resources remain optional user resources, while the module still requires its workload operator to be available.

## Validation

```bash
go test ./...
```

All tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added operator health monitoring before the module reports readiness.
  * Operator availability, deployment failures, and missing deployments now produce corresponding health statuses and reasons.

* **Bug Fixes**
  * Enabled services with no operand instances are now reported as healthy.
  * Modules with no default workloads, or enabled services without workloads, now correctly report readiness.


kpunwatk (approved):

ruivieira:
@coderabbitai can you open a follow up issue so that the Module smoke test Kind also includes de workload operator?

